PR #3, opened 2022-09-04
Safe access to nilable error

bhelx:
There appears to be a situation in which this err can be nil


nilslice (approved):
Good catch! thanks 🙏

PR #4, opened 2022-09-04
Avoid indexing zero-length slice in go

zshipko:


nilslice:
tested in context of https://github.com/extism/extism-fsnotify/pull/2 and it indeed solves the issue. thank you! will merge once CI is done.

nilslice (approved):

PR #5, opened 2022-09-04
Create Extism module / namespace

bhelx:
I noticed when calling set_log_file that it's on the top-level namespace.
And so is the rest of the gem. It's not a good idea to do this as it
pollutes the user's application. Best practice is to wrap all code in a
module which by convention is the gem name.

I was also getting `Uninitialized constant FFI:NIL` and in the
set_log_file method and I could not tell why it was needed. I changed it
to a regular ruby nil. If it's needed for some reason I'll change it
back or move to another PR.


ruby/example.rb
```diff
@@ -4,7 +4,7 @@
 manifest = {
   :wasm => [{:path => "../wasm/code.wasm"}]
 }
-plugin = Plugin.new(manifest)
+plugin = Extism::Plugin.new(manifest)
```
bhelx, line 7: Perhaps the example in the documentation site needs to be changed as well? or is this links to that automatically?

ruby/lib/extism.rb
```diff
@@ -1,57 +1,59 @@
...
   end
-  C.extism_log_file(name, level)
-end
-
 
-class Plugin
-  def initialize(wasm, wasi=false, config=nil)
-    if wasm.class == Hash then
-      wasm = JSON.generate(wasm)
+  def self.set_log_file(name, level=nil)
```
bhelx, line 20: here i replaced `FFI:NIL` with `nil`

nilslice:
great call, thanks! @zshipko, I think this should be fine -- but maybe give it a quick look before merge. I can handle updating the docs site if you give this a +1

nilslice:
wondering why the action isnt running.. would like to see tests pass before merge. maybe the conflicts in `ruby/lib/extism.rb` are blocking it?

zshipko:
@nilslice I think the action isn't running because of the merge conflict - other than that it looks good!

bhelx:
Was missing some changes from my remote's main

nilslice (approved):

PR #6, opened 2022-09-06
feat: Add extism_plugin_update

zshipko:
This gives host the ability to re-use plugin descriptors instead of loading a new plugin each time. The plugin memory and everything is reset, so no state is shared with the newly loaded plugin. This could be used to assign a static plugin ID to a specific user or to reload an updated WASM file, I think there are a lot of ways this could come in handy.

I tried to update every SDK, but couldn't get PHP to work with my environment - @nilslice maybe you could take a look at that one?


nilslice:
Cool! Yes, I'll take a look at it this evening and push to this branch.

runtime/src/sdk.rs
```diff
@@ -29,13 +29,43 @@ pub unsafe extern "C" fn extism_plugin_register(
...
 
+#[no_mangle]
+pub unsafe extern "C" fn extism_plugin_update(
+    index: PluginIndex,
+    wasm: *const u8,
+    wasm_size: Size,
+    with_wasi: bool,
+) -> bool {
+    let index = index as usize;
+    trace!("Call to extism_plugin_replace with wasm pointer {:?}", wasm);
```
nilslice, line 46:
```suggestion
    trace!("Call to extism_plugin_update with wasm pointer {:?}", wasm);
```

nilslice: although, "replace" is maybe more correct? since `Plugin::new` pretty much wipes away everything at the original plugin descriptor.

runtime/src/sdk.rs
```diff
@@ -29,13 +29,43 @@ pub unsafe extern "C" fn extism_plugin_register(
...
+    let mut plugins = match PLUGINS.lock() {
+        Ok(p) => p,
+        Err(e) => e.into_inner(),
+    };
+
+    if index < plugins.len() {
+        plugins[index] = plugin;
+    }
+
+    info!("Plugin replaced: {index}");
```
nilslice, line 65:
```suggestion
    info!("Plugin updated: {index}");
```

runtime/src/sdk.rs
zshipko: Yeah I was on the fence about the naming, I still think update is a little more descriptive in the context of updating an existing plugin with a new version, but replace would also work.

nilslice: yea I like them both, no issue either way. update also opens up more things like update_manifest - maybe eventually other fields on the `Plugin` that make sense to be updated.

nilslice:
Just a quick update.. I am working on the PHP sdk, just a bit slowly. Blocked on trying to understand the class autoloading stuff that composer sets up & I've not really got much experience with. But will figure it out!

nilslice:
@zshipko - I think this is good to merge, and want to get the PHP sdk deployed to packagist asap.. So, once CI passes (I added PHP test back in), I'll merge.

nilslice (approved):

PR #7, opened 2022-09-09
fix: use empty allocation when no config key is present

nilslice:
@zshipko - pretty sure this is not optimal, but idk if there is a path to the same effect without actually calling the alloc routines. 

re: https://github.com/extism/assemblyscript-pdk/pull/4


zshipko:
I've opened a PR with another solution here: https://github.com/extism/extism/pull/9

nilslice:
Closing in favor of #9

PR #9, opened 2022-09-09
Handle missing config keys without raising an error

zshipko:
- Makes it so `extism_length(0)` returns 0, as far as I can tell this seems safe
- Returns 0 when a key can't be found


nilslice:
Nice - ok, I'll try it out & merge once CI passes - thanks!

nilslice (approved):

PR #16, opened 2022-09-14
Add `scripts` directory with SDK coverage script

zshipko:
- Adds `scripts/header.py` to parse `runtime/extism.h` into a Python class
- Adds `scripts/sdk_coverage.py` to check if functions defined in `runtime/extism.h` are used in each SDK
  - Depends on `pycparser` from pip and `ripgrep`
- Removes `runtime/extism.json` because it doesn't match the header file anymore and I couldn't figure out how it was generated

Here is the current output:
```
go	88.89%	(8/9)
	1. extism_function_exists

haskell	100.00%	(9/9)

node	100.00%	(9/9)

ocaml	88.89%	(8/9)
	1. extism_function_exists

php	77.78%	(7/9)
	1. extism_function_exists
	2. extism_log_file

python	88.89%	(8/9)
	1. extism_function_exists

ruby	88.89%	(8/9)
	1. extism_function_exists

rust	100.00%	(9/9)
```


nilslice:
This is great!!

> Removes runtime/extism.json because it doesn't match the header file anymore and I couldn't figure out how it was generated

Ah, I shouldn't have committed that without the change to build.rs 

For posterity, in case it's useful in the future, I generated it by calling to clang with -ast-dump=json on the header file, and then used https://github.com/dtolnay/clang-ast to extract the parts I was interested in using to generate the coverage check.

nilslice:
Now that I've made a complete mess of this PR, (trying to add a comment from CI w/ the coverage report lol)... I'll merge once this passes. Decided to balk on the PR comment, and simply add a workflow job to generate the report for each run. I'll revisit this in the future, but hit my time budget.

nilslice (approved):

PR #17, opened 2022-09-16
feat: update node sdk to use async call 

Jhaines1988:


nilslice:
Awesome! Thank you for adding this. I'll wait for the rest of the tests to pass CI and merge after that.

nilslice (approved):
🚀 ✅ 🏆

PR #18, opened 2022-09-17
Implement Host SDK for Elixir

bhelx:
This implements the a basic Elixir / Erlang host SDK. I still have some work to do on it this weekend.

## Testing

* Pull down this branch
* In a different folder create a new elixir project:
  `mix new extism_elixir_host && cd extism_elixir_host`
* Link to the local dependency by updating the `deps` in `mix.exs`:
```elixir
  defp deps do
    [
      {:extism, path: "/Users/ben/Code/extism/elixir"}
    ]
  end
```
* Run `mix deps.get`
* Run `mix deps.compile`, this should trigger the rust compilation in the elixir lib depdency
* Run `iex -S mix` to start a REPL.
* Test happy path:
```elixir
iex(1)> manifest = %{ wasm: [ %{ path: "/Users/ben/code/extism/wasm/code.wasm" } ]}
%{wasm: [%{path: "/Users/ben/code/extism/wasm/code.wasm"}]}
iex(2)> {:ok, plugin} = Extism.Plugin.new(manifest)
{:ok,
 %Extism.Plugin{
   resource: 0,
   reference: #Reference<0.1101693245.999030787.163527>
 }}
iex(3)> {:ok, result} = Extism.Plugin.call(plugin, "count_vowels", "this is a test")
{:ok, "{\"count\": 4}"}
```
* Test the sad path:
```elixir
iex(1)> manifest = %{ wasm: [ %{ path: "/Users/ben/code/extism/wasm/code.wasmz" } ]}
%{wasm: [%{path: "/Users/ben/code/extism/wasm/code.wasmz"}]}
iex(2)> Extism.Plugin.new(manifest) 
{:error, "Could not load plugin"}
iex(3)> manifest = %{ wasm: [ %{ pathe: "/Users/ben/code/extism/wasm/code.wasm" } ]}
%{wasm: [%{pathe: "/Users/ben/code/extism/wasm/code.wasm"}]}
iex(4)> Extism.Plugin.new(manifest)                                                 
{:error, "Could not parse manifest"}
```
* Some other functions you can use:
```elixir
iex(8)> Extism.Plugin.set_log_file(plugin, "/tmp/mylog.log", "debug")  
:ok
iex(9)> Extism.Plugin.update(plugin, %{wasm: [%{path: "/Users/ben/Code/extism/wasm/code.wasm"}]}, true)
{:ok, true}
iex(10)> Extism.Plugin.has_function(plugin, "unknown")
false
iex(11)> Extism.Plugin.has_function(plugin, "count_vowels")
true
```

## Running unit tests

```
cd elixir/
mix do deps.get, deps.compile
mix test
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,8 @@ jobs:
   build_and_test:
     name: Build & Test
     runs-on: ${{ matrix.os }}
+    env:
+      MIX_ENV: test
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
@@ -44,6 +46,20 @@ jobs:
       - name: Install extism shared library
         shell: bash
         run: sudo make install
+
+      - name: Setup Elixir Host SDK
+        if: ${{ runner.os != 'macOS' }}
+        uses: erlef/setup-beam@v1
+        with:
+          experimental-otp: true
+          otp-version: '25.0.4'
+          elixir-version: '1.14.0'
+
+      - name: Test Elixir Host SDK
+        if: ${{ runner.os != 'macOS' }}
+        run: |
+         cd elixir
+         LD_LIBRARY_PATH=/usr/local/lib mix do deps.get, test
 
       - name: Setup Go env
         uses: actions/setup-go@v3
--- a/elixir/lib/extism/plugin.ex
+++ b/elixir/lib/extism/plugin.ex
@@ -1,0 +1,45 @@
+defmodule Extism.Plugin do
+  defstruct [
+    # The actual NIF Resource. PluginIndex and the context
+    plugin_id: nil,
+    ctx: nil
+  ]
+
+  def wrap_resource(ctx, plugin_id) do
+    %__MODULE__{
+      ctx: ctx,
+      plugin_id: plugin_id
+    }
+  end
+
+  def call(plugin, name, input) do
+    case Extism.Native.plugin_call(plugin.ctx.ptr, plugin.plugin_id, name, input) do
+      {:error, err} -> {:error, err}
+      res -> {:ok, res}
+    end
+  end
+
+   def update(plugin, manifest, wasi) when is_map(manifest) do
+     {:ok, manifest_payload} = JSON.encode(manifest)
+     case Extism.Native.plugin_update_manifest(plugin.ctx.ptr, plugin.plugin_id, manifest_payload, wasi) do
+       {:error, err} -> {:error, err}
+       res -> :ok
+     end
+   end
+
+   def free(plugin) do
+     Extism.Native.plugin_free(plugin.ctx.ptr, plugin.plugin_id)
+   end
+
+   def has_function(plugin, function_name) do
+     Extism.Native.plugin_has_function(plugin.ctx.ptr, plugin.plugin_id, function_name)
+   end
+end
+
+defimpl Inspect, for: Extim.Plugin do
+  import Inspect.Algebra
+
+  def inspect(dict, opts) do
+    concat(["#Extism.Plugin<", to_doc(dict.plugin_id, opts), ">"])
+  end
+end
--- a/elixir/native/extism_nif/Cargo.toml
+++ b/elixir/native/extism_nif/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "extism_nif"
+version = "0.0.1-rc.3"
+edition = "2021"
+authors = ["Benjamin Eckel <bhelx@simst.im>"]
+
+[lib]
+name = "extism_nif"
+path = "src/lib.rs"
+crate-type = ["cdylib"]
+
+[dependencies]
+rustler = "0.26.0"
+extism = { version = "0.0.1-rc.3", path = "../../../rust" }
+log = "0.4"
+
--- a/elixir/native/extism_nif/src/lib.rs
+++ b/elixir/native/extism_nif/src/lib.rs
@@ -1,0 +1,149 @@
+use rustler::{Atom, Env, Term, ResourceArc};
+use extism::{Plugin, Context};
+use std::str;
+use std::path::Path;
+use std::str::FromStr;
+use std::sync::RwLock;
+use std::mem;
+
+mod atoms {
+    rustler::atoms! {
+        ok,
+        error,
+        unknown // Other error
+    }
+}
+
+struct ExtismContext {
+    ctx: RwLock<Context>
+}
+
+fn load(env: Env, _: Term) -> bool {
+    rustler::resource!(ExtismContext, env);
+    true
+}
+
+fn to_rustler_error(extism_error: extism::Error) -> rustler::Error {
+    match extism_error {
+        extism::Error::UnableToLoadPlugin(msg) => rustler::Error::Term(Box::new(msg)),
+        extism::Error::Message(msg) => rustler::Error::Term(Box::new(msg)),
+        extism::Error::Json(json_err) => rustler::Error::Term(Box::new(json_err.to_string()))
+    }
+}
+
+#[rustler::nif]
+fn context_new() -> ResourceArc<ExtismContext> {
+    ResourceArc::new(
+        ExtismContext { ctx: RwLock::new(Context::new()) }
+    )
+}
+
+#[rustler::nif]
+fn context_reset(ctx: ResourceArc<ExtismContext>) {
+    let context = &mut ctx.ctx.write().unwrap();
+    context.reset()
+}
+
+#[rustler::nif]
+fn context_free(ctx: ResourceArc<ExtismContext>) {
+    let context = &ctx.ctx.read().unwrap();
+    std::mem::drop(context)
+}
+
+#[rustler::nif]
+fn plugin_new_with_manifest(ctx: ResourceArc<ExtismContext>, manifest_payload: String, wasi: bool) -> Result<i32, rustler::Error> {
+    let context = &ctx.ctx.write().unwrap();
+    let result = match Plugin::new(context, manifest_payload, wasi) {
+        Err(e) => Err(to_rustler_error(e)),
+        Ok(plugin) => {
+            let plugin_id = plugin.as_i32();
+            // this forget should be safe because the context will clean up
+            // all it's plugins when it is dropped
+            mem::forget(plugin);
+            Ok(plugin_id)
+        }
+    };
+    result
+}
+
+#[rustler::nif]
+fn plugin_call(ctx: ResourceArc<ExtismContext>, plugin_id: i32, name: String, input: String) -> Result<String, rustler::Error> {
+    let context = &ctx.ctx.read().unwrap();
+    let plugin = unsafe { Plugin::from_id(plugin_id, context) };
+    let result = match plugin.call(name, input) {
+        Err(e) => Err(to_rustler_error(e)),
+        Ok(result) => {
+            match str::from_utf8(&result) {
+                Ok(output) => Ok(output.to_string()),
+                Err(_e) => Err(rustler::Error::Term(Box::new("Could not read output from plugin")))
+            }
+        }
+    };
+    // this forget should be safe because the context will clean up
+    // all it's plugins when it is dropped
+    mem::forget(plugin);
+    result
+}
+
+#[rustler::nif]
+fn plugin_update_manifest(ctx: ResourceArc<ExtismContext>, plugin_id: i32, manifest_payload: String, wasi: bool) -> Result<(), rustler::Error> {
+    let context = &ctx.ctx.read().unwrap();
+    let mut plugin = unsafe { Plugin::from_id(plugin_id, context) };
+    let result = match plugin.update(manifest_payload, wasi) {
+        Ok(()) => {
+            Ok(())
+        },
+        Err(e) => Err(to_rustler_error(e))
+    };
+    // this forget should be safe because the context will clean up
+    // all it's plugins when it is dropped
+    mem::forget(plugin);
+    result
+}
+
+#[rustler::nif]
+fn plugin_free(ctx: ResourceArc<ExtismContext>, plugin_id: i32) -> Result<(), rustler::Error> {
+    let context = &ctx.ctx.read().unwrap();
+    let plugin = unsafe { Plugin::from_id(plugin_id, context) };
+    std::mem::drop(plugin);
+    Ok(())
+}
+
+#[rustler::nif]
+fn set_log_file(filename: String, log_level: String) -> Result<Atom, rustler::Error> {
+    let path = Path::new(&filename);
+    match log::Level::from_str(&log_level) {
+        Err(_e) => Err(rustler::Error::Term(Box::new(format!("{} not a valid log level", log_level)))),
+        Ok(level) => {
+            extism::set_log_file(path, Some(level));
+            Ok(atoms::ok())
+        }
+    }
+}
+
+#[rustler::nif]
+fn plugin_has_function(ctx: ResourceArc<ExtismContext>, plugin_id: i32, function_name: String) -> Result<bool, rustler::Error> {
+    let context = &ctx.ctx.read().unwrap();
+    let plugin = unsafe { Plugin::from_id(plugin_id, context) };
+    let has_function = plugin.has_function(function_name);
+    // this forget should be safe because the context will clean up
+    // all it's plugins when it is dropped
+    mem::forget(plugin);
+    Ok(has_function)
+}
+
+rustler::init!(
+    "Elixir.Extism.Native",
+    [
+        context_new,
+        context_reset,
+        context_free,
+        plugin_new_with_manifest,
+        plugin_call,
+        plugin_update_manifest,
+        plugin_has_function,
+        plugin_free,
+        set_log_file,
+    ],
+    load = load
+);
--- a/elixir/test/extism_test.exs
+++ b/elixir/test/extism_test.exs
@@ -1,0 +1,81 @@
+defmodule ExtismTest do
+  use ExUnit.Case
+  doctest Extism
+
+  test "context create & reset" do
+    ctx = Extism.Context.new()
+    path = Path.join([__DIR__, "../../wasm/code.wasm"])
+    manifest = %{wasm: [%{path: path}]}
+    {:ok, plugin} = Extism.Context.new_plugin(ctx, manifest, false)
+    Extism.Context.reset(ctx)
+    # we should expect an error after resetting context
+    {:error, _err} = Extism.Plugin.call(plugin, "count_vowels", "this is a test")
+  end
+
+  defp new_plugin() do
+    ctx = Extism.Context.new()
+    path = Path.join([__DIR__, "../../wasm/code.wasm"])
+    manifest = %{wasm: [%{path: path}]}
+    {:ok, plugin} = Extism.Context.new_plugin(ctx, manifest, false)
+    {ctx, plugin}
+  end
+
+  test "counts vowels" do
+    {ctx, plugin} = new_plugin()
+    {:ok, output} = Extism.Plugin.call(plugin, "count_vowels", "this is a test")
+    assert JSON.decode(output) == {:ok, %{"count" => 4}}
+    Extism.Context.free(ctx)
+  end
+
+  test "can make multiple calls on a plugin" do
+    {ctx, plugin} = new_plugin()
+    {:ok, output} = Extism.Plugin.call(plugin, "count_vowels", "this is a test")
+    assert JSON.decode(output) == {:ok, %{"count" => 4}}
+    {:ok, output} = Extism.Plugin.call(plugin, "count_vowels", "this is a test again")
+    assert JSON.decode(output) == {:ok, %{"count" => 7}}
+    {:ok, output} = Extism.Plugin.call(plugin, "count_vowels", "this is a test thrice")
+    assert JSON.decode(output) == {:ok, %{"count" => 6}}
+    Extism.Context.free(ctx)
+  end
+
+  test "can free a plugin" do
+    {ctx, plugin} = new_plugin()
+    {:ok, output} = Extism.Plugin.call(plugin, "count_vowels", "this is a test")
+    assert JSON.decode(output) == {:ok, %{"count" => 4}}
+    Extism.Plugin.free(plugin)
+    # Expect an error when calling a plugin that was freed
+    {:error, _err} = Extism.Plugin.call(plugin, "count_vowels", "this is a test")
+    Extism.Context.free(ctx)
+  end
+
+  test "can update manifest" do
+    {ctx, plugin} = new_plugin()
+    path = Path.join([__DIR__, "../../wasm/code.wasm"])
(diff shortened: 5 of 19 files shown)


elixir/native/extism_nif/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+                Err(_e) => Err(Error::Term(Box::new("Could not load plugin"))),
+                Ok(plugin) => Ok(plugin.as_isize())
+            }
+        },
+        Err(_e) => Err(Error::Term(Box::new("Could not parse manifest")))
+    }
+}
+
+ #[rustler::nif]
+fn call_plugin(plugin_iz: isize, name: String, input: String) -> Result<String, Error> {
```
zshipko, line 33: Thanks for working on this - it looks great!

My one early bit of feedback is the other SDKs don't require input/output to be UTF8 compatible (`Vec<u8>` instead of `String` here) - I'm not opposed to making an exception if you think this is a nicer API for elixir or if there's some other reason this works better.

rust/src/lib.rs
```diff
@@ -6,7 +6,13 @@ use extism_manifest::Manifest;
...
 
 #[repr(transparent)]
-pub struct Plugin(isize);
+pub struct Plugin(pub isize);
+
+impl Plugin {
+    pub fn as_isize(self) -> isize {
+        self.0
+    }
+}
```
bhelx, line 15: It's been a long time since I've written rust so I ran into a few obstacles. I added this so I could go from the "newtype" `Plugin` back to `isize`. See the `plugin_new_with_manifest` implementation where I cast back to isize which is automatically mapped to erlang space. I believe there is a way to implement a rustler specific trait for this type but i couldn't do it when the type is defined outside of the crate. Maybe you have a simpler solution.

elixir/native/extism_nif/src/lib.rs
bhelx: Let me see how ergonomic it is on the erlang side. We have a wrapper around the `Native` module that it exposes so we can always transform types coming in and out if we need.

elixir/lib/extism/plugin.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  def wrap_resource(resource) do
+    %__MODULE__{
+      resource: resource,
+      reference: make_ref()
+    }
+  end
+
+  def new(manifest, wasi \\ false) when is_map(manifest) do
+    {:ok, manifest_payload} = JSON.encode(manifest)
```
bhelx, line 21: Instead of sharing the Manifest struct, which rustler does allow with some magic, I decided to just encode a map to json and send it as a string. I'm open to other ways of doing this. One downside with this is you need the validation in rust land and you would probably catch less problems at elixir compile time. The upside is less interface coupling to this language.

elixir/lib/extism/plugin.ex
```diff
@@ -0,0 +1,58 @@
...
+    # resource in attributes. This will convert the resource into an
+    # empty binary with no warning. This will make that harder to
+    # accidentaly do.
+    # It also serves as a handy way to tell plugins apart.
+    reference: nil
+  ]
+
+  def wrap_resource(resource) do
+    %__MODULE__{
+      resource: resource,
```
bhelx, line 15: `resource` is just the PluginIndex here. The elixir user sees the `%Plugin` type as the reference. We may not need the `reference` property here I just cargo culted it.

elixir/lib/extism/plugin.ex
```diff
@@ -0,0 +1,42 @@
...
+      res -> {:ok, res}
+    end
+  end
+end
+
+defimpl Inspect, for: Extim.Plugin do
+  import Inspect.Algebra
+
+  def inspect(dict, opts) do
+    concat(["#Extism.Plugin<", to_doc(dict.reference, opts), ">"])
```
nilslice, line 56: I wonder if `resource` is better to print here on `inspect` -- this formatting is maybe used for error/log messages? 

As you mentioned above, its the `PluginIndex`, which is referenced a lot in the runtime logs. So using it here might also help tie Elixir logs w/ runtime logs.

rust/src/lib.rs
nilslice: that works! I think it would be better to use the `From` trait here and implement it for both `Plugin` and `isize` bi-directionally. I have a commit to your branch incoming.

.github/workflows/ci.yml
```diff
@@ -45,6 +45,19 @@ jobs:
...
+          experimental-otp: true
+          otp-version: '25.0'
+          elixir-version: '1.14.0'
+
+      - name: Test Elixir Host SDK
+        run: |
+          cd elixir
+          mix deps.get
+          MIX_ENV=test mix do compile, test
+
```
bhelx, line 60: Works on my machine but can't seem to find the `.so` on github actions.

elixir/lib/extism/plugin.ex
bhelx: Yeah, I might get rid of reference too since the index should be unique per instance, correct?

elixir/native/extism_nif/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+// fn load(env: Env, _: Term) -> bool {
+//     true
+// }
+
+ #[rustler::nif]
+fn plugin_new_with_manifest(manifest_payload: String, wasi: bool) -> Result<isize, Error> {
+    match serde_json::from_str(&manifest_payload) {
```
bhelx, line 23: Instead of parsing should I just pass the rendered json right down to the SDK?

elixir/native/extism_nif/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+
+ #[rustler::nif]
+fn call_plugin(plugin_iz: isize, name: String, input: String) -> Result<String, Error> {
+    let plugin = Plugin(plugin_iz);
+    match plugin.call(name, input) {
+        Err(_e) => Err(Error::Term(Box::new("Failed to call plugin"))),
+        Ok(result) => {
+            match str::from_utf8(&result) {
+                Ok(output) => Ok(output.to_string()),
+                Err(_e) => Err(Error::Term(Box::new("Could not read output from plugin")))
```
bhelx, line 42: Going to come back with some better error handling soon

elixir/native/extism_nif/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+// fn load(env: Env, _: Term) -> bool {
+//     true
+// }
+
+ #[rustler::nif]
+fn plugin_new_with_manifest(manifest_payload: String, wasi: bool) -> Result<isize, Error> {
+    match serde_json::from_str(&manifest_payload) {
+        Ok(manifest) => {
+            match Plugin::new_with_manifest(&manifest, wasi) {
+                Err(_e) => Err(Error::Term(Box::new("Could not load plugin"))),
```
bhelx, line 26: I'm going to come back with some better error handling soon. This is just for development purposes.

nilslice:
> Test the sad path:
iex(1)> manifest = %{ wasm: [ %{ path: "/Users/ben/code/extism/wasm/code.wasmz" } ]}
%{wasm: [%{path: "/Users/ben/code/extism/wasm/code.wasmz"}]}
iex(2)> Extism.Plugin.new(manifest) 
{:error, "Could not load plugin"}

To clarify,

In this test, is `/Users/ben/code/extism/wasm/code.wasmz` a valid path with a wasm binary at its location? or is this a non-existent file intentionally used here to fail? 

In either case, we could probably be more specific in the error message, e.g. `{:error, "Could not load plugin: no such file or directory" }` or `{:error, "Could not load plugin: failed to parse manifest" }` etc.

.github/workflows/ci.yml
nilslice: Hm, yea this works locally for me as well. I can look into it further later this weekend!

elixir/native/extism_nif/src/lib.rs
nilslice: Yea I'd go straight to the SDK. The `Result` still can be handled nicely in Elixir I think, which is returned from the `Plugin::new_with_manifest` anyways.

elixir/lib/extism/plugin.ex
nilslice: Ok, thanks for calling this out! Let me think about it and test around a bit

ruby/lib/extism/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Extism
-  VERSION = "0.1.0"
+  VERSION = "0.0.1.rc1"
```
bhelx, line 4: Sorry: drive by commit in the ruby library

elixir/lib/extism/plugin.ex
```diff
@@ -0,0 +1,51 @@
...
+  def has_function(plugin, function_name) do
+    Extism.Native.plugin_has_function(plugin.plugin_index, function_name)
+  end
+end
+
+defimpl Inspect, for: Extim.Plugin do
+  import Inspect.Algebra
+
+  def inspect(dict, opts) do
+    concat(["#Extism.Plugin<", to_doc(dict.plugin_index, opts), ">"])
```
bhelx, line 49: I changed this to just be the plugin_index and changed the name to help with debugging:

```elixir
iex(1)> {:ok, plugin} = Extism.Plugin.new(%{wasm: [%{path: "/Users/ben/Code/extism/wasm/code.wasm"}]})
{:ok, %Extism.Plugin{plugin_index: 0}}
iex(2)> {:ok, plugin} = Extism.Plugin.new(%{wasm: [%{path: "/Users/ben/Code/extism/wasm/code.wasm"}]})
{:ok, %Extism.Plugin{plugin_index: 1}}
iex(3)> {:ok, plugin} = Extism.Plugin.new(%{wasm: [%{path: "/Users/ben/Code/extism/wasm/code.wasm"}]})
{:ok, %Extism.Plugin{plugin_index: 2}}
iex(4)> {:ok, plugin} = Extism.Plugin.new(%{wasm: [%{path: "/Users/ben/Code/extism/wasm/code.wasm"}]})
{:ok, %Extism.Plugin{plugin_index: 3}}
```

elixir/native/extism_nif/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+authors = ["Benjamin Eckel <bhelx@simst.im>"]
+
+[lib]
+name = "extism_nif"
+path = "src/lib.rs"
+crate-type = ["cdylib"]
+
+[dependencies]
+rustler = "0.25.0"
+extism = { version = "0.0.1-alpha", path = "../../../rust" }
```
bhelx, line 14: This makes sense for development and testing, but probably not for release. Probably want:

```toml
extism = { version = "0.0.1-rc.3" }
```

Brings up the same convo about which version of the SDK the clients are using I think.

bhelx: Perhaps we could conditionally swap out for development / test / and release. We could use a matrix to test both local and crates version.

nilslice:
This is looking great, @bhelx. Thanks a ton!

I can help with CI, and just to remind you, I've made a PR against your forked branch to add the `From` impl so the `Plugin(isize)`'s wrapped `isize` can stay private. Would you take a look at that & merge if you agree with the approach? https://github.com/bhelx/extism/pull/1

rust/src/lib.rs
```diff
@@ -8,6 +8,18 @@ mod bindings;
 #[repr(transparent)]
 pub struct Plugin(isize);
 
+impl From<isize> for Plugin {
+    fn from(i: isize) -> Self {
+        Plugin(i)
+    }
+}
+
+impl From<Plugin> for isize {
```
bhelx, line 17: @nilslice I had to implement From going the other way for `into()` to work. does this seem right?

```
error[E0277]: the trait bound `isize: From<Plugin>` is not satisfied
  --> src/lib.rs:31:33
   |
31 |         Ok(plugin) => Ok(plugin.into())
   |                                 ^^^^ the trait `From<Plugin>` is not implemented for `isize`
   |
   = help: the following other types implement trait `From<T>`:
             <f32 as From<i16>>
             <f32 as From<i8>>
             <f32 as From<u16>>
             <f32 as From<u8>>
             <f64 as From<f32>>
             <f64 as From<i16>>
             <f64 as From<i32>>
             <f64 as From<i8>>
           and 67 others
   = note: required because of the requirements on the impl of `Into<isize>` for `Plugin`
```

elixir/native/extism_nif/Cargo.toml
bhelx: We could also release possibly with a vendored version of the library.

bhelx: Curious what your thoughts are @nilslice

rust/src/lib.rs
```diff
@@ -8,6 +8,18 @@ mod bindings;
...
+    fn from(p: Plugin) -> Self {
+        p.0
```
zshipko, line 19: This doesn't actually work because `drop` is implemented for `Plugin` so it won't live beyond this function and the ID returned now pointed to a deallocated plugin.

The plugin lifetime probably needs to be managed manually using `std::mem::forget` and `drop` - or we could add a function to prevent `Plugin` from being freed on `drop`

elixir/native/extism_nif/Cargo.toml
nilslice: `extism = { version = "0.0.1-rc.3", path = "../../../rust" }` effectively means, "use `version` from crates.io, unless `path` is found". Where `path` is prioritized as the source of the crate when developing locally.

Not sure if that addresses your point above, but happy to dig in more when we can sync up!

rust/src/lib.rs
nilslice: Yea, sorry if I may have done the reverse in the other PR... You can also use `From::from(plugin)` if only one is available, but I personally tend to prefer using `.into()` when I can as you've done here.

elixir/lib/extism/plugin.ex
nilslice: I think this is probably a fine solution for the time being - especially if we think the `Manifest` type itself will be in flux until a more stable version of Extism in the future. 

It's probably worth exploring how we'd define this type in `witx` format, and try to use (and contribute to) a tool such as: 

https://github.com/bytecodealliance/wit-bindgen or https://github.com/jedisct1/witx-codegen

elixir/test/extism_test.exs
```diff
@@ -0,0 +1,54 @@
...
+
+  test "counts vowels" do
+    Extism.set_log_file("/tmp/extism.log", "DEBUG")
+    ctx = Extism.Context.new()
+    path = Path.join([__DIR__, "../../wasm/code.wasm"])
+    manifest = %{wasm: [%{path: path}]}
+    {:ok, plugin} = Extism.Context.new_plugin(ctx, manifest, false)
+    {:ok, plugin} = Extism.Context.new_plugin(ctx, manifest, false)
+    {:ok, plugin} = Extism.Context.new_plugin(ctx, manifest, false)
+    {:ok, output} = Extism.Plugin.call(plugin, "count_vowels", "this is a test")
```
bhelx, line 26: Tried creating a few plugins and saw the plugin_id increment. But when trying to call i get the plugin with id doesn't exist panic.

rust/src/context.rs
```diff
@@ -18,11 +18,14 @@ impl Context {
...
 
     /// Remove all registered plugins
-    pub fn reset(&mut self) {
+    pub fn reset(&self) {
         unsafe { bindings::extism_context_reset(self.pointer) }
     }
 }
 
+unsafe impl Send for Context {}
+unsafe impl Sync for Context {}
```
bhelx, line 27: Rustler asked me to add this

rust/src/plugin.rs
```diff
@@ -6,6 +6,17 @@ pub struct Plugin<'a> {
...
+    pub fn from(id: i32, context: &'a Context) -> Plugin<'a> {
+        Plugin {
+            id,
+            context: context,
+        }
+    }
+
+    pub fn as_i32(&self) -> i32 {
+        self.id
+    }
```
bhelx, line 18: I needed to add ways to go to and from the primitive values in the rust code in `extism_nif/src/lib.rs`

bhelx: I think these can go away if i change to pointer as zach suggested.

rust/src/context.rs
```diff
@@ -18,11 +18,14 @@ impl Context {
     }
 
     /// Remove all registered plugins
-    pub fn reset(&mut self) {
```
bhelx, line 21: I don't think this mut was needed

rust/src/context.rs
nilslice: I don't think we currently don't cross thread boundaries, but if we do in the future (a multi-threaded execution env might be necessary at some point), this change means that we need to be aware of synchronizing plugin access. 

These are "marker traits" that tell the type system if any value is OK to be shared across threads. So by adding these, Rust will just take our word that there are no memory issues / data races. But, there could be. @zshipko - would be good to get your insight on this & any potential impacts, or preventative measures we should take if needed.

zshipko: Yeah, I'm not entirely sure what the consequences of adding `Send` and `Sync` here are - if we wanted to make the context struct threadsafe (which we probably do) we could wrap the plugin map in the mutex.

.github/workflows/ci.yml
```diff
@@ -93,6 +95,21 @@ jobs:
...
+          experimental-otp: true
+          otp-version: '25.0.4'
+          elixir-version: '1.14.0'
+
+      - name: Test Elixir Host SDK
+        run: |
+         cd elixir &&
+         mix deps.get &&
+         mix deps.compile &&
+         mix compile &&
```
nilslice, line 110: Github actions is set up to fail on a per-line basis here, so I think you can leave out the `&&`. Curious though if you tried that and it didn't work!

rust/src/context.rs
nilslice: maybe it was there to symbolically indicate that the operation mutates an underlying resource, and if we only want that to happen when the resource is owned in the scope its mutated, then the `mut` keyword will prevent Rust from compiling anything that breaks this requirement. @zshipko?

zshipko: Yeah the intent was to prevent `reset` from being called when there are multiple references

.github/workflows/ci.yml
bhelx: Yeah, i did this back when i was debugging something. Will change it back.

rust/src/context.rs
bhelx: Okay, I'm just realizing this was not a drive by commit. Needing a mutable reference adds some complexity around the calling code. @zshipko Should I make the Context mutable and wrap in a mutex?

rust/src/context.rs
bhelx: I double checked and rustler just won't work without these traits. Maybe there is a way i'm missing.

```
[I] ben@m1lap ~/C/e/elixir (elixir-nif)> mix test
Compiling 1 file (.ex)
Compiling crate extism_nif in debug mode (native/extism_nif)
   Compiling extism v0.0.1-rc.3 (/Users/ben/Code/extism/rust)
   Compiling extism_nif v0.0.1-rc.3 (/Users/ben/Code/extism/elixir/native/extism_nif)
error[E0277]: `*mut extism::bindings::ExtismContext` cannot be sent between threads safely
  --> src/lib.rs:22:5
   |
22 |     rustler::resource!(ExtismContext, env);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `*mut extism::bindings::ExtismContext` cannot be sent between threads safely
   |
   = help: within `extism::Context`, the trait `Send` is not implemented for `*mut extism::bindings::ExtismContext`
   = note: required because it appears within the type `extism::Context`
   = note: required because of the requirements on the impl of `Sync` for `RwLock<extism::Context>`
note: required because it appears within the type `ExtismContext`
  --> src/lib.rs:17:8
   |
17 | struct ExtismContext {
   |        ^^^^^^^^^^^^^
note: required by a bound in `ResourceTypeProvider`
  --> /Users/ben/.cargo/registry/src/github.com-1ecc6299db9ec823/rustler-0.26.0/src/resource.rs:37:48
   |
37 | pub trait ResourceTypeProvider: Sized + Send + Sync + 'static {
   |                                                ^^^^ required by this bound in `ResourceTypeProvider`
   = note: this error originates in the macro `rustler::resource` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: `*mut extism::bindings::ExtismContext` cannot be shared between threads safely
  --> src/lib.rs:22:5
   |
22 |     rustler::resource!(ExtismContext, env);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `*mut extism::bindings::ExtismContext` cannot be shared between threads safely
   |
   = help: within `extism::Context`, the trait `Sync` is not implemented for `*mut extism::bindings::ExtismContext`
   = note: required because it appears within the type `extism::Context`
   = note: required because of the requirements on the impl of `Sync` for `RwLock<extism::Context>`
note: required because it appears within the type `ExtismContext`
  --> src/lib.rs:17:8
   |
17 | struct ExtismContext {
   |        ^^^^^^^^^^^^^
note: required by a bound in `ResourceTypeProvider`
  --> /Users/ben/.cargo/registry/src/github.com-1ecc6299db9ec823/rustler-0.26.0/src/resource.rs:37:48
   |
37 | pub trait ResourceTypeProvider: Sized + Send + Sync + 'static {
   |                                                ^^^^ required by this bound in `ResourceTypeProvider`
   = note: this error originates in the macro `rustler::resource` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `extism_nif` due to 2 previous errors

== Compilation error in file lib/extism/native.ex ==
** (RuntimeError) Rust NIF compile error (rustc exit code 101)
    (rustler 0.26.0) lib/rustler/compiler.ex:41: Rustler.Compiler.compile_crate/2
    lib/extism/native.ex:2: (module)
```


I'll see if i can find a way around using the `resource!` macro. Maybe i can somehow pass the raw pointer without upsetting rust.

.github/workflows/ci.yml
```diff
@@ -45,6 +47,20 @@ jobs:
         shell: bash
         run: sudo make install
 
+      - name: Setup Elixir Host SDK
+        if: ${{ runner.os != 'macOS' }}
```
bhelx, line 51: This action is not supported on macOS yet. I'm testing on mac so it should be okay for now.

rust/src/plugin.rs
```diff
@@ -6,6 +6,17 @@ pub struct Plugin<'a> {
 }
 
 impl<'a> Plugin<'a> {
+    pub unsafe fn from(id: i32, context: &'a Context) -> Plugin<'a> {
```
zshipko, line 9: If this is going to be public, could you rename it to `Plugin::from_id` or something more specific?

PR #19, opened 2022-09-17
Add ExtismContext to SDK + better errors for failed register/update

zshipko:
- Adds `ExtismContext` instead of global `PLUGINS` registry
   - Adds `extism_context_new`, `extism_context_free` and `extism_context_reset`
   - Requires updating nearly every SDK function to add context parameter
- Renames some SDK functions to follow better naming conventions
   - `extism_plugin_register` -> `extism_plugin_new`
   - `extism_output_get` -> `extism_plugin_output_data`
   - `extism_output_length` -> `extism_plugin_output_length`
   - `extism_call` -> `extism_plugin_call`
- Updates `extism_error` to return the context error when -1 issued for the plug-in ID
- Adds `extism_plugin_free` to remove an existing plugin
- Updates SDKs to include these functions
  - It looks like only Rust, OCaml, Python, PHP, C++ have the ability to write destructors, while Haskell, Ruby, Go, C and Node will require users to call the `free` method - do we want to keep it like this? I think it's nice to take advantage of the automatic cleanup when possible, but also see the value in having the SDKs match.
- 100% coverage for all SDKs (except C++, since it can use C functions directly for some operations)


rust/src/lib.rs
```diff
@@ -37,26 +43,40 @@ impl Plugin {
...
+            let buf = buf.to_str().unwrap().to_string();
+            return Err(Error::UnableToLoadPlugin(buf));
         }
 
         Ok(Plugin(plugin as isize))
     }
 
-    pub fn update(&mut self, data: impl AsRef<[u8]>, wasi: bool) -> bool {
-        unsafe {
+    pub fn update(&mut self, data: impl AsRef<[u8]>, wasi: bool) -> Result<bool, Error> {
```
bhelx, line 55: 🥳

nilslice:
> It looks like only Rust, OCaml, Python, PHP, C++ have the ability to write destructors [...]

While I agree it's nice to keep the SDKs consistent with each other, in this case I'd personally prefer to leverage all the features in any respective language. So using the destructor options where they are available is maybe more idiomatic. 

Thoughts?

ruby/lib/extism.rb
```diff
@@ -74,5 +91,9 @@ def call(name, data, &block)
       buf = C.extism_output_get(@plugin)
       return block.call(buf, out_len)
     end
+    
+    def destroy()
+      C.extism_plugin_destroy(@plugin)
+    end
```
bhelx, line 103: This comment applies to any languages with GC, but since we initialize a plugin when we initialize the object, should we also destroy the plugin when the object gets freed by the GC? In ruby these are called "finalizers". In python i think there is also a way to hook into this. We could also use context managers.

bhelx: The more I think about it, the more I think it's maybe a bad idea. Because there could be cases where the GC collects an object but the plugin isn't done working and it would be a very tricky thing to debug. It's probably better to keep this explicit and just make sure that people understand that they are allocating this plugin and may need to deallocate it themselves explicitly if they are creating many of them.

zshipko: I wasn’t able to find information about finalizers in Ruby. But I think it would be ideal to use finalizers when possible and if we find out there is some kind of lifetime issue we can fix that - I don’t think it will be an issue in practice.

In Python we’re using the __del__ method to do the cleanup.

bhelx: > What is the typical way of doing finalization in Ruby?

The semantics for this in ruby a little weird. I see in python that the GC eventually will call `__del__` for you if the reference count goes to 0.
In ruby you can define a finalizer in the initializer:

```ruby
class Plugin  
  def initialize(wasm, wasi=false, config=nil)
    # ...
    ObjectSpace.define_finalizer(self, proc { |_id| self.destroy } )
  end
end
```

> But I think it would be ideal to use finalizers when possible and if we find out there is some kind of lifetime issue we can fix that - I don’t think it will be an issue in practice.

My concern would be someone say creating a plugin and firing off a call to it from a web-request. then the plugin object goes out of scope and gets garbage collected when the request is done (but maybe the plugin is not done). It does seem like there is no asynchronous way to do a call though. So maybe that's not a problem since call has to return (meaning work is done). I've yet to observe or test this just thinking out loud.

zshipko: Ah that is an interesting point, I wasn’t thinking about the possibility of async plugins. On the Rust side the plug-in is protected by a mutex so I don’t think that can actually happen. Either way that can be something we handle in the Rust code.

I will try adding that to the Ruby SDK, thanks for the example!

zshipko:
I added finalizers for ruby and node, but they aren't exactly reliable. The docs for both mention that the callback may never be called.

We should probably just mention somewhere that if you want to *ensure* a plugin is destroyed you need to manually call the `destroy` method (or equivalent)

ruby/lib/extism.rb
zshipko: I kept getting warnings about the finalizer referencing the object being finalized, but I found a slightly more complex way that works using a global hash from object_ids to plugin ids

runtime/src/sdk.rs
```diff
@@ -113,6 +113,20 @@ pub unsafe extern "C" fn extism_plugin_destroy(plugin: PluginIndex) {
...
+        // Stop at first `Some`
+        if x.is_some() {
+            if i < plugins.len() {
+                plugins.truncate(i)
+            }
+
+            break;
+        }
+    }
 }
```
nilslice, line 130: Nice! Before we merge this, I may add some tests. Now that there is some destructive behavior I think it would be good to have a few checks around what is destroyed/retained and verify that there are no side effects.

zshipko:
@nilslice, `ExtismContext` has been added and all the SDKs are updated (my original comment is also updated to describe the new updates) - I will give it a once over tomorrow and it would be helpful if you could do the same. After this is merged I think it's probably best to start working on using some more of the host functionality in the tests to make sure things are working correctly.

As usual, I'm having some issues with the php SDK, I feel like I may have messed something up. Any help figuring out what's going on there is appreciated, but I will keep poking at it for a little.

nilslice:
Nice! I'll definitely do some testing, and 100% agree it's time to step up the host SDK test coverage. I'll plan to do some work in that department over the next couple of days. This looks like an awesome improvement by the way. 

cc/ @bhelx to get some early eyes on it when you get a moment - especially as to how it impacts your elixir PR.

nilslice:
@zshipko - just finished reviewing the runtime/rust changes and added a few comments and trace logs. Great job on this, I think it's a much more complete package & usable in bigger/sophisticated software projects.

I'll continue the review of the SDKs, and test the fsnotify example after I update it to use these changes.

python/extism/extism.py
```diff
@@ -105,47 +107,107 @@ def _wasm(plugin):
     return wasm
 
 
+class Context:
```
bhelx, line 110: Minor note:

Perhaps this could also support the context manager dunder methods since it semantically maps well to the `with` language feature: https://docs.python.org/3/library/contextlib.html

bhelx:
> cc/ @bhelx to get some early eyes on it when you get a moment - especially as to how it impacts your elixir PR.

It certainly affects the Elixir API but I can quickly refactor to come up with a context equivalent. So if you want to merge before mine go ahead.

python/example.py
```diff
@@ -15,11 +15,11 @@
 hash = hashlib.sha256(wasm).hexdigest()
 config = {"wasm": [{"data": wasm, "hash": hash}], "memory": {"max": 5}}
 
-plugin = Plugin(config)
-
-# Call `count_vowels`
-j = json.loads(plugin.call("count_vowels", data))
-print("Number of vowels:", j["count"])
+with Context() as context:
+    with context.plugin(config) as plugin:
```
bhelx, line 19: My assumption was that the context_free call would also free up any active plugins in memory. Is that the case? If so you could just use the Context block:

```python
with Context() as context:
  plugin = context.plugin(config)
  j = json.loads(plugin.call("count_vowels", data))
  print("Number of vowels:", j["count"])
```

And when it exits it would free up any plugins you created

Would be the same with the other libraries. For example you'd only need to `defer` the context in go.

runtime/src/sdk.rs
```diff
@@ -5,74 +5,135 @@ use std::str::FromStr;
...
+}
+
+/// Free a context
+#[no_mangle]
+pub unsafe extern "C" fn extism_context_free(ctx: *mut Context) {
+    trace!("Freeing context");
+    if ctx.is_null() {
+        return;
+    }
+    drop(Box::from_raw(ctx))
```
bhelx, line 22: Should we iterate the plugins here and free them? It would prevent the progammers from needing to track and clean up the plugins they make in it's scope.

python/example.py
zshipko: True, good point! I will update the example

runtime/src/sdk.rs
zshipko: I believe this does clean up all the plugins automatically since they’re just references to values stored in the `plugins` field. The only reason `extism_plugin_free` exists is if you specifically want the context to outlive the plugin. But now that the context stuff exists we can de-emphasize the need to free plugins.

bhelx: I see, makes sense! I need to re-read about rust lifetimes.

python/example.py
```diff
@@ -11,11 +11,13 @@
...
+# a Context provides a scope for plugins to be managed within. creating multiple contexts
+# is expected and groups plugins based on source/tenant/lifetime etc.
 with Context() as context:
+    wasm = open("../wasm/code.wasm", 'rb').read()
+    hash = hashlib.sha256(wasm).hexdigest()
+    config = {"wasm": [{"data": wasm, "hash": hash}], "memory": {"max": 5}}
+
     plugin = context.plugin(config)
     # Call `count_vowels`
     j = json.loads(plugin.call("count_vowels", data))
```
nilslice, line 23: @zshipko - to me, it makes more sense to demo the code this way so that it's more apparent that a Context should live longer than a plugin (or span across multiple plugins) vs. a single context along with a plugin. Not saying that was the intent of the code previously, but putting the context creation _after_ the plugin configuration made it appear that the context is unique to a single plugin. 

I could have this wrong though, so please let me know if you think otherwise!

nilslice:
Ok, I think the SDKs look great and I've updated the fsnotify example here https://github.com/extism/extism-fsnotify/pull/3/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261 (didn't take much, despite the significant underlying change! 👍)

I'll wait for CI to complete on this PR and then I think it's good to merge as long as you don't have any additional changes @zshipko?

zshipko:
I don’t have any more changes, so it seems ready to merge.

And I agree, we should definitely add some better documentation to each of the SDKs to explain how they each work.

nilslice (approved):
:shipit:

PR #22, opened 2022-09-23
Don't panic on invalid plugin id

zshipko:


bhelx (approved):
Much appreciated! If possible we should probably ban panics in the runtime.

runtime/src/export.rs
```diff
@@ -1,6 +1,22 @@
 /// All the functions in the file are exposed from inside WASM plugins
 use crate::*;
 
+// This macro unwraps input arguments to prevent functions from panicking,
+// it should be used instead of `Val::unwrap_*` functions
+macro_rules! get {
```
nilslice, line 6: nit: maybe a more descriptive name here: `get_val`? `get_offset`?

runtime/src/context.rs
```diff
@@ -57,6 +57,7 @@ impl Context {
 
     /// Set the context error
     pub fn set_error(&mut self, e: impl std::fmt::Debug) {
+        trace!("Set context error: {:?}", e);
```
nilslice, line 60: would it make sense to use `error!` here? or do we likely want to catch that at the source of the error instead (and  using `trace!` here is best for the intentionally verbose debugging level?)

nilslice (approved):
Great as-is, just a couple questions/nits. If you think it's good to go, merge away!

runtime/src/context.rs
zshipko: I believe the error is being logged somewhere else, this log was added while debugging some issues with the error message itself.

PR #23, opened 2022-09-23
Add `allowed_hosts` configuration option

zshipko:
- Adds `allowed_hosts` to the manifest
- If there are any `allowed_hosts` then `extism_http_request` checks to make sure a URL's host is listed before performing the request
 
One thing to note: I have a branch that also adds `allowed_dirs`, but I'm not sure if that's something I want to include. We had kinda talked about not supporting filesystem access, which simplifies everything. The reason I ended up making that branch is when experimenting with `python.wasm` it seemed to require the current directory to be preopened. Since Python plugin support is a long ways off and we can probably hack around the filesystem requirement it's probably best to leave it `allowed_dirs` out until we know if we'll need it.

### DIFF
--- a/extism.go
+++ b/extism.go
@@ -67,7 +67,8 @@ type Manifest struct {
 	Memory struct {
 		Max uint32 `json:"max,omitempty"`
 	} `json:"memory,omitempty"`
-	Config map[string]string `json:"config,omitempty"`
+	Config       map[string]string `json:"config,omitempty"`
+	AllowedHosts []string          `json:"allowed_hosts,omitempty"`
 }
 
 func makePointer(data []byte) unsafe.Pointer {
--- a/haskell/src/Extism/Manifest.hs
+++ b/haskell/src/Extism/Manifest.hs
@@ -13,6 +13,7 @@ valueOrNull f Nothing = JSNull
 valueOrNull f (Just x) = f x
 makeString s = JSString (toJSString s)
 stringOrNull = valueOrNull makeString
+makeArray f [] = JSNull
 makeArray f x = JSArray [f a | a <- x]
 filterNulls obj = [(a, b) | (a, b) <- obj, not (isNull b)]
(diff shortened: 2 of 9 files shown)


nilslice:
Thank you for the quick turnaround here - especially if it helps unblock the user in Discord!

One consideration to add now (if it's straightforward) would be to allow a glob pattern to match? So I could do `*.extism.org` or similar variants of this, which I think would be pretty common use case. 

A question for you (which may be better addressed in another issue/PR) is if/how this could be applied to WASI as well, not specifically `http_request`?

zshipko:
I added the glob matching like you mentioned, it wasn't a big change.

Right now there isn't a good way to make this work with WASI directly, I think we would need to use this: https://github.com/deislabs/wasi-experimental-http or wait for the WASI sockets proposal to be implemented. Unfortunately that repo doesn't seem to work with wasmtime 1.0 yet so I haven't been able to test it out. In the meantime maybe we can do something with wit-bindgen to make `http_request` accessible to more languages.

We should also figure out how much WASI we want to expose and then add those configurations in another PR. Do we want optional filesystem access? Do we want to make it an option to use stdin/stderr/stdout? I am on the fence about that stuff, it might be interesting to see how far we can go without enabling that stuff, but at the same time being able to write to a file or stdout would be super convenient.

nilslice (approved):

nilslice:
> In the meantime maybe we can do something with wit-bindgen to make http_request accessible to more languages.

Yea, this would also help to start push us in the direction of the Component Model, which is a good thing to inch towards.

> We should also figure out how much WASI we want to expose and then add those configurations in another PR. Do we want optional filesystem access? Do we want to make it an option to use stdin/stderr/stdout? I am on the fence about that stuff, it might be interesting to see how far we can go without enabling that stuff, but at the same time being able to write to a file or stdout would be super convenient.

Let's definitely plan on having a call about this, maybe inviting the Discord to an open conversation about it sometime next week or so.

PR #30, opened 2022-10-17
Add `extism_version` to Host SDK and all clients

bhelx:
Adds a new Host SDK call `extism_version` which returns the version string.

### DIFF
--- a/runtime/extism.h
+++ b/runtime/extism.h
@@ -107,4 +107,7 @@ const uint8_t *extism_plugin_output_data(struct ExtismContext *ctx, ExtismPlugin
  */
 bool extism_log_file(const char *filename, const char *log_level);
 
-const uint8_t *extism_version(void);
+/**
+ * Get the Extism version string
+ */
+const char *extism_version(void);
--- a/runtime/extism.h
+++ b/runtime/extism.h
@@ -107,4 +107,7 @@ const uint8_t *extism_plugin_output_data(struct ExtismContext *ctx, ExtismPlugin
  */
 bool extism_log_file(const char *filename, const char *log_level);
 
+/**
+ * Get the Extism version string
+ */
 const char *extism_version(void);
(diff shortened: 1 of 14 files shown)


runtime/src/sdk.rs
```diff
@@ -385,3 +385,8 @@ pub unsafe extern "C" fn extism_log_file(
     }
     true
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn extism_version() -> *const u8 {
+    env!("CARGO_PKG_VERSION").as_ptr()
```
bhelx, line 391: Is this the right manifest to grab the version from? it seems we are on release `v0.0.1-rc.4` but this one is showing `rc.3`.

bhelx: My mistake. We identified my main branch was pointed to the wrong origin.

zshipko:
Good idea! it looks like Go and Haskell need to be updated too

runtime/extism.h
```diff
@@ -106,3 +106,5 @@ const uint8_t *extism_plugin_output_data(struct ExtismContext *ctx, ExtismPlugin
  * Set log file and level
  */
 bool extism_log_file(const char *filename, const char *log_level);
+
+const uint8_t *extism_version(void);
```
bhelx: This appears to be generated but i'm not sure from where. missing a docstring

zshipko: I think converting this comment to a doc comment should copy the documentation over: https://github.com/extism/extism/pull/30/files#diff-9c18909a3b1c4dcdf57dc4edc707e98ba17728203f7c17d573a5d1de60ba9095R389

zshipko: (It's generated from sdk.rs using cbindgen in the build script)

bhelx: yes, appears i just annotated that comment incorrectly

zshipko (approved):

PR #33, opened 2022-10-20
feat: Add workflow_dispatch to the release workflow

bhelx:


.github/workflows/release.yml
```diff
@@ -1,4 +1,6 @@
 on:
+  workflow_dispatch:
+    workflow: release-sdks
```
bhelx, line 3: gonna see if this even works

PR #35, opened 2022-10-20
Fix Elixir cargo dependency

bhelx:
I thought this would look for the local path then fallback to the crates.io version. Ideally we want this to be conditional somehow.


bhelx:
I'm gonna move forward with this one since it fixes the problem for users. then i will come back and re-visit the other problems.

PR #39, opened 2022-10-24
feat(elixir):  use result of set_log_file

bhelx:
We can merge this after we release `rc.6`


bhelx:
This is mergable now that rc.6 is released

PR #46, opened 2022-10-25
test(elixir-sdk): Add Elixir to SDK coverage script

bhelx:


zshipko:
The way the SDK coverage script works is it checks for each function listed in `extism.h` in the source of each SDK,  since the elixir bindings go through Rust the SDK coverage script won't work. I will try to work on an update at some point.

bhelx:
There should be matching elixir code in this native.ex file https://github.com/extism/extism/blob/main/elixir/lib/extism/native.ex

Does that help?

zshipko:
Actually, after looking at it again, it will still fail (even with the right path) because those functions are missing the `extism_` prefix - we could update the sdk coverage script to check with and without the prefix.

nilslice:
@zshipko - good catch... 

It looks like maybe we should combine these files to scan (not sure if that's ok in Elixir):
https://github.com/extism/extism/blob/main/elixir/lib/extism/context.ex 
https://github.com/extism/extism/blob/main/elixir/lib/extism/plugin.ex

Since https://github.com/extism/extism/blob/main/elixir/lib/extism/native.ex doesn't actually _call_ the runtime APIs. 

If the `elixir_nif` code was updated to include the `extism_` prefixes on the Rust functions, and rebuilt/generated, I think the declaration code would have them. Then the Elixir `lib` stuff above would need to be updated to call the prefixed functions.

bhelx:
Closing for the time being

PR #61, opened 2022-11-02
chore: update deps, add dependabot.yml

zshipko:


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "cargo" # See documentation for possible values
```
nilslice, line 8: Since we also host all the SDKs in this repo, do you think it would be good to add ecosystem settings for the other languages as well? I know we don't use a many libraries in the SDKs now, but that could change in the future?

zshipko: Sure, I will look into that

zshipko: Alright, `dependabot.yml` has been updated for the sdks languages they support - take a look to make sure it seems right.

.github/dependabot.yml
```diff
@@ -0,0 +1,41 @@
...
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
```
nilslice, line 11: Just double-checking, this means that dependabot is workspace-aware - right?

zshipko: Based on this: https://github.com/dependabot/dependabot-core/issues/1207#issuecomment-500635306 - I think it is.

nilslice (approved):
This will be super helpful!

PR #65, opened 2022-11-03
Make Rust SDK depend directly on `extism-runtime`

zshipko:


rust/src/context.rs
```diff
@@ -1,8 +1,6 @@
 use crate::*;
 
-pub struct Context {
-    pub(crate) pointer: *mut bindings::ExtismContext,
-}
+pub struct Context(pub(crate) std::cell::RefCell<extism_runtime::Context>);
```
nilslice, line 3: Can you share a bit about using this smart pointer here vs. the raw pointer used previously? If we were to consider sharing plugin contexts across threads (or in something like an async task), how does this impact the changes needed - or are they mostly the same as with the raw pointer.

zshipko: This was meant to mimic the behavior of the raw pointer before, otherwise certain functions would have to be converted to `&mut self` which would cause issues since `&'a Context` is embedded in `Plugin` and there's no good way to get a mutable reference. 

It should be no problem to update this to work with a `Mutex` instead of a `RefCell` if you think multithreading issues might come up down the road.

nilslice: Ok I gotcha - thanks!

PR #66, opened 2022-11-04
docs: Link to the manifest concept doc

bhelx:


extism.go
```diff
@@ -194,15 +194,15 @@ func (plugin Plugin) SetConfig(data map[string][]byte) error {
 	return nil
 }
 
-/// FunctionExists returns true when the name function is present in the plugin
+// / FunctionExists returns true when the name function is present in the plugin
```
zshipko, line 197: Is this extra slash a mistake? it kind of looks like I may have used Rust's doc comments when writing comments for Go and the formatter didn't like it so it changed it to this.

nilslice, line 197:
```suggestion
// FunctionExists returns true when the named function is present in the plugin
```

nilslice (dismissed):
Pending the edits / suggestions, LGTM!

extism.go
```diff
@@ -194,15 +194,15 @@ func (plugin Plugin) SetConfig(data map[string][]byte) error {
...
 func (plugin Plugin) FunctionExists(functionName string) bool {
 	name := C.CString(functionName)
 	b := C.extism_plugin_function_exists(plugin.ctx.pointer, C.int(plugin.id), name)
 	C.free(unsafe.Pointer(name))
 	return bool(b)
 }
 
-/// Call a function by name with the given input, returning the output
+// / Call a function by name with the given input, returning the output
```
nilslice, line 205:
```suggestion
// Call a function by name with the given input, returning the output
```

extism.go
bhelx: Interesting. maybe my editor did that

PR #67, opened 2022-11-04
Bump to version 0.0.1-rc.6

bhelx:


runtime/Cargo.toml
```diff
@@ -28,7 +28,7 @@ log4rs = "1.1"
 url = "2.3"
 glob = "0.3"
 ureq = {version = "2.5", optional=true}
-extism-manifest = { version = "0.0.1-rc.4", path = "../manifest" }
+extism-manifest = { version = "0.0.1-rc.5", path = "../manifest" }
```
zshipko, line 31:
```suggestion
extism-manifest = { version = "0.0.1-rc.6", path = "../manifest" }
```

ruby/lib/extism/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Extism
-  VERSION = "0.0.1.rc4"
+  VERSION = "0.0.1.rc6"
```
nilslice, line 4: Is it an Elixir specific thing to exclude the `.` between rc and the number?

I see we already released it without last time - so nbd, since we'll be off this version soon enough. Just curious.

nilslice (approved):

ruby/lib/extism/version.rb
bhelx: https://guides.rubygems.org/patterns/

it's how they recommend it. Either way may work though. Im not sure

PR #68, opened 2022-11-04
fix: Fix release workflow to use make

bhelx:


.github/workflows/release.yml
```diff
@@ -237,7 +237,7 @@ jobs:
           CI: true
         run: |
           cd node
-          npm publish
+          make publish
```
bhelx, line 240: make publish basically does `clean -> build -> release`

PR #69, opened 2022-11-04
ci: Workflow per SDK

bhelx:


nilslice:
nit: updated the PR title (thus the commit message) to be `ci` prefixed

.github/workflows/release.yml
```diff
@@ -1,8 +1,7 @@
 on:
-  workflow_dispatch:
-    workflow: release-sdks
```
bhelx, line 3: I thought this would allow me to run this action only but didn't realize it had dependency on the runtime release

.github/workflows/release-rust.yaml
```diff
@@ -0,0 +1,35 @@
...
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_API_TOKEN }}
+        run: |
+          # order of crate publication matter: manifest, runtime, rust
+
+          cargo publish --manifest-path manifest/Cargo.toml
+          # allow for crates.io to update so dependant crates can locate extism-manifest
+          sleep 5 
+
+          cargo publish --manifest-path runtime/Cargo.toml --no-verify
+          cargo publish --manifest-path rust/Cargo.toml
```
nilslice, line 33: eventually we may need to split these 3 out as well -- since these will probably be on their own release cycles.

nilslice (approved):

PR #70, opened 2022-11-05
feat: [Experimental] Extism browser runtime

bhelx:
Experimental browser runtime. Will allow you to load and run simple Extism plugins in the browser.

Still a work-in-progress. I've currently implemented just enough to run the `count_vowels` plugin. To try it out:

```
cd browser
npm install
npm run build
python3 -m http.server
```

Open http://localhost:8000

There is a small playground style app. Set the url, and the function name. Paste in some input. Hit `Run`. The result of the plugin call should appear on the right.

Screen Shot 2022-11-05 at 2 01 24 PM


browser/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+        this.input = input
+        let func = this.module?.instance.exports[func_name]
+        if (!func){
+            throw Error(`function does not exist ${func_name}`)
+        }
+        //@ts-ignore
+        func()
+        let output = `${this.output}`
+        this.output = undefined
+        this.input = undefined
```
bhelx, line 100: This instance variable setting and clearing is a little awkward. Might want to instantiate the module per call and pass in another closure with fresh input and output variables.

bhelx: Same goes with memory too i guess. the whole scope needs to be separate from the plugin instance.

browser/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    //@ts-ignore
+    return new Uint8Array(new BigInt64Array([n]).buffer)
+}
+
+function packU8toU64(bytes: Uint8Array): BigInt {
+    return (new BigInt64Array(bytes.buffer))[0]
+}
+
+function stringToBytes(s: string, offset: number, len: number): Uint8Array {
+    return new Uint8Array(s.slice(offset, len).split("").map(c => c.charCodeAt(0)))
```
bhelx, line 12: probably a much better way to do this

browser/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+            const i = Number(idx)
+            //@ts-ignore
+            return packU8toU64(stringToBytes(plugin.input, i, i+8))
+        },
+        extism_output_set(offset: BigInt, len: number): number {
+            //@ts-ignore
+            offset = Number(offset)
+            len = Number(len)
+            //@ts-ignore
+            plugin.output = String.fromCharCode(...plugin.memory.slice(offset, offset+len))
```
bhelx, line 53: we probably want to keep the output and input as raw bytes and let the programmer decide if they represent strings or not

browser/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    return (new BigInt64Array(bytes.buffer))[0]
+}
+
+function stringToBytes(s: string, offset: number, len: number): Uint8Array {
+    return new Uint8Array(s.slice(offset, len).split("").map(c => c.charCodeAt(0)))
+}
+
+function makeEnv(plugin: ExtismPlugin): any {
+    return {
+        extism_alloc(n: BigInt): BigInt {
```
zshipko, line 17: This only works if the plugin only uses extism_alloc once, I think the http example might fail. I can help port some of the allocation logic over from the Rust runtime.

browser/src/index.test.ts
```diff
@@ -0,0 +1,6 @@
+import { sum } from "./";
+describe("sum function", () => {
+  it("adds 2 numbers", () => {
+    expect(sum(1, 2)).toBe(3);
+  });
```
bhelx, line 5: I'll write some tests once we agree on an interface

browser/src/index.ts
```diff
@@ -0,0 +1,103 @@
+
+function unpackU64toU8(n: BigInt): Uint8Array {
+    //@ts-ignore
+    return new Uint8Array(new BigInt64Array([n]).buffer)
+}
+
+function packU8toU64(bytes: Uint8Array): BigInt {
+    return (new BigInt64Array(bytes.buffer))[0]
```
bhelx, line 8: maybe a better way to pack and unpack bytes. But my bit shifting and mask code was buggy and I think in danger of overflows. I assume this is safe but could use some scrutiny

browser/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+        extism_store_u8(offset: BigInt, n: number) {
+            //@ts-ignore
+            plugin.memory[offset] = Number(n)
+        },
+        extism_store_u32(n: BigInt, i: number) {debugger; },
+        extism_store_u64(offset: BigInt, n: BigInt) {
+            let parts = unpackU64toU8(n)
+            parts.forEach((p, i) => {
+                plugin.memory[i + Number(offset)] = p
+            })
```
zshipko, line 33: I think you could do something like:

```typescript
   let cast = new BigUint64Array(plugin.memory.buffer, offset);
   cast[0] = n;
```

browser/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+            return BigInt(plugin.input.length)
+        },
+        extism_input_load_u8(i: number): number {
+            //@ts-ignore
+            return plugin.input[i].charCodeAt(0)
+        },
+        extism_input_load_u64(idx: BigInt) {
+            const i = Number(idx)
+            //@ts-ignore
+            return packU8toU64(stringToBytes(plugin.input, i, i+8))
```
zshipko, line 46: Similar to above:

```typescript
   let cast = new BigUint64Array(plugin.memory.buffer, Number(idx));
   return cast[0];
```

bhelx:
Still a work in progress. Will bring subsequent PRs before we release

PR #74, opened 2022-11-10
chore: bump browser runtime to rc.10 and refactor

bhelx:


browser/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "@extism/runtime-browser",
-  "version": "0.0.1-rc.6",
+  "version": "0.0.1-rc.10",
   "description": "Extism runtime in the browser",
   "scripts": {
-    "build": "node build.js",
+    "build": "node build.js && tsc --emitDeclarationOnly --outDir dist",
```
bhelx, line 6: `npm-dts` was not generating valid type defs. this is the best hack i could come up with. let tsc do it instead of relying on esbuild. I could maybe implement this in build.js before release.

nilslice: if it works, it works ;) thank you

nilslice (approved):

nilslice:
oops i didnt notice at first, but looks like an import syntax update is needed: 
https://github.com/extism/extism/actions/runs/3440098498/jobs/5738183996#step:6:28

PR #76, opened 2022-11-14
fix: Fix the python release workflow

bhelx:


bhelx:
Getting failure to reupload here which means it worked: https://github.com/extism/extism/actions/runs/3465513773/jobs/5788329117

PR #77, opened 2022-11-14
chore: Fix elixir release workflow

bhelx:


bhelx:
Kind of hacky, but seems to be working now: https://github.com/extism/extism/actions/runs/3472085858/jobs/5802438473

failing to republish the same version

nilslice:
I see -- so the publish step doesn't need the underlying .so so it still triggers this warning: 

```
Warning: {:error,
 {:load_failed,
  'Failed to load NIF library /home/runner/work/extism/extism/elixir/_build/dev/lib/extism/priv/native/libextism_nif: \'libextism.so: cannot open shared object file: No such file or directory\''}}
```

.github/workflows/release-elixir.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Download libextism
         uses: actions/download-artifact@v3
         with:
-          name: libextism-${{ matrix.os }}
+          name: libextism-ubuntu-latest
```
nilslice: `${{ matrix.os }}` is only valid if there is a strategy matrix defined in `job`: 
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

nilslice: I think it would still evaluate to an empty string though?

.github/workflows/release-elixir.yaml
```diff
@@ -18,6 +18,8 @@ jobs:
       - name: Install extism shared library
         shell: bash
         run: |
+          mkdir -p /home/runner/.local/bin/  
+          export PATH="/home/runner/.local/bin/ :$PATH"
```
nilslice, line 22: Does the extra space bewteen `/bin/` and `:$PATH` exist in the resulting variable? Worried that would cause an issue.

bhelx: no i'll fix though

.github/workflows/release-elixir.yaml
bhelx: I think i deleted this

PR #81, opened 2022-11-18
feat: add `extism_http_status_code` to get the status code of the last HTTP request + fixes for clippy

zshipko:
This is kind of hacky but until we have an established way of encoding complex types in memory it seems good enough


nilslice:
Nice! We can also always wrap this up behind some nice library API in the PDKs.

nilslice (approved):

PR #82, opened 2022-11-20
fix(node-sdk): Properly calculate length of utf-8 inputs

bhelx:
Properly calculate the length of utf-8 inputs. I'll work on getting unicode chars into our other tests as well and see if we need to fix any other SDKs

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -94,6 +94,12 @@ if (process.env.EXTISM_PATH) {
 
 const lib = locate(searchPath);
 
+// get the true byte size of a utf-8 string
+const byteSize = (data: string | Buffer) => {
+  if (Buffer.isBuffer(data)) return data.length;
+  return new Blob([data]).size;
+}
+
 /**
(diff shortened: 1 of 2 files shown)


node/src/index.ts
```diff
@@ -94,6 +94,12 @@ if (process.env.EXTISM_PATH) {
 
 const lib = locate(searchPath);
 
+// get the true byte size of a utf-8 string
+const byteSize = (data: string | Buffer) => {
```
bhelx, line 98: perhaps instead of doing this we can just always turn the string into a buffer. There might be a better way to do this.

zshipko: It looks like there is also `Buffer.byteLength`: https://nodejs.org/api/buffer.html#static-method-bufferbytelengthstring-encoding

bhelx: Thanks i'll try that

bhelx: Seems to work!

nilslice (approved):

PR #87, opened 2022-11-21
feat: add var/config functions, merge ExtismPlugin and ExtismPluginCall

zshipko:
This also only instantiates the module once and combines `ExtismPlugin` with `ExtismPluginCall`, I can pull out just the new function implementations if that's an issue.

### DIFF
--- a/browser/src/plugin.ts
+++ b/browser/src/plugin.ts
@@ -19,12 +19,24 @@ export default class ExtismPlugin {
     this.output = new Uint8Array();
   }
 
-  async getExportedFunctions() {
+  async getExports(): Promise<WebAssembly.Exports> {
     const module = await this._instantiateModule();
-    return Object.keys(module.instance.exports).filter((f) => !f.startsWith('__') && f !== 'memory');
+    return module.instance.exports;
+  }
+
+  async getImports(): Promise<WebAssembly.ModuleImportDescriptor[]> {
+    const module = await this._instantiateModule();
+    return WebAssembly.Module.imports(module.module);
+  }
+
+  async getInstance(): Promise<WebAssembly.Instance> {
+    const module = await this._instantiateModule();
+    return module.instance;
   }
 
   async call(func_name: string, input: Uint8Array | string): Promise<Uint8Array> {
+    const module = await this._instantiateModule();
+
     if (typeof input === 'string') {
       this.input = new TextEncoder().encode(input);
     } else if (input instanceof Uint8Array) {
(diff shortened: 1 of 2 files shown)


browser/src/plugin.ts
```diff
@@ -1,15 +1,61 @@
...
+    this.allocator = new Allocator(1024 * 1024);
+    this.config = config;
+    this.vars = {};
+    this.input = new Uint8Array();
+    this.output = new Uint8Array();
+  }
+
+  async getExportedFunctions() {
+    const module = await this._instantiateModule();
+    return Object.keys(module.instance.exports).filter((f) => !f.startsWith('__') && f !== 'memory');
```
nilslice, line 24: curious about the `__` prefix filter, do we really want to exclude these? I get it for the use-case in the Playground, but do we need it here? 

maybe it would be better to add a single getter for the instance stuff (imports/exports/etc).

nilslice:
@zshipko - nice work! how does someone use the var/config stuff though? do we need to add some additional methods to call the inner functions that expect `bigint` args?

zshipko:
Thanks! the config can be set in the plugin constructor function and the vars are only accessed using `extism_var_get`/`extism_var_set` - the bigint parameters are equivalent to `uint64_t` in C world.

nilslice:
:man_facepalming: i was totally thinking of this as a PDK for a moment... guess I need more coffee

nilslice:
will leave review/approval for @bhelx who is more familiar with the usage of this currently. LGTM tho!

browser/src/plugin.ts
```diff
@@ -24,20 +70,13 @@ class ExtismPluginCall {
       extism_load_u8(n: bigint): number {
         return plugin.allocator.memory[Number(n)];
       },
-      extism_load_u32(n: bigint): number {
```
bhelx, line 27: Should we leave these but just throw an exception? what happens if you use a plugin that expects them to be there?

nilslice: there shouldnt be any plugins that expect these -- weve also removed them from all other sdks so this is consistent with the rest of the world

browser/src/plugin.ts
bhelx: Yeah agreed. If this is copied from me, I used this to make the dropdown for the playground.

bhelx: It might make more sense to just give them all the exports

zshipko: I will update it - should we still filter out the `memory` export?

nilslice: personally, i'd go the route of adding getters for the things on `module.instance` we want to make available. not explicitly supporting a way to list the functions here. let a user of this library make that decision. 

so like on an instantiated `Plugin`, we could add class methods to `getImports()` and `getExports()`, which then could be used in whatever way a caller wants?

browser/src/plugin.ts
zshipko: yeah, these functions were removed but also have never been used in any of the PDKs, so it's unlikely to find a plugin that uses them.

bhelx: IIRC, the older plugins expect the whole interface to be satisfied even though they don't use them

zshipko: i think that's only true if the function is actually called somewhere

bhelx (approved):

PR #95, opened 2022-11-28
fix: resolve php sdk issues

nilslice:
There is still an issue somewhere which causes the content of `extism.h` to be printed to stdout. Unsure why...


bhelx (approved):

kmapb:
fwiw invoking the php interpreter on a file that lacks a `<?` glyph in it will print that file. If `extsim.h` were somehow either in the autoload map or require'd from a PHP source file I'd expect to see that behavior.

nilslice:
@kmapb that must be it - autoloader probably reads all the included files in the composer.json - one of which is indeed extism.h

great tip! I'll try this out

nilslice:
@kmapb - removing the header file from the composer.json lost fixed the issue. Thanks again!

PR #103, opened 2022-11-29
chore: Add Makefile for browser runtime

bhelx:
I accidentally published without building. this should help make sure releases are valid

### DIFF
--- a/browser/Makefile
+++ b/browser/Makefile
@@ -1,0 +1,25 @@
+.PHONY: test
+
+prepare:
+	npm install
+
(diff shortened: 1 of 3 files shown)


nilslice:
Thanks! Is the idea that these will eventually be run in the CI and Release workflows?

bhelx:
> Thanks! Is the idea that these will eventually be run in the CI and Release workflows?

For now i'm doing most of the released locally. I just want to use a makefile to it ensures things are properly built when publishing

PR #105, opened 2022-11-29
fix(rust-sdk): remove runtime publish from rust release script

bhelx:


nilslice:
Is this still being published elsewhere? If it isn't published I think the Rust SDK will break.

zshipko:
Yeah, it looks like `extism-runtime` isn't being published anywhere else, so this might not be the right change. 

It might make the most sense to create a separate release action for each of the crates? Or is that too much?

bhelx:
closing since it's stale

PR #108, opened 2022-11-29
docs(elixir-sdk): Reduce elixir min version

bhelx:
I've tested it in at least as low as 1.12. It just prints out a warning so should be fine if someone wants to use it on an older version.


nilslice:
Good call!

Let's wait on merging this until the Rust/rustler issue is sorted out so we can be sure the Elixir workflow passes.

PR #110, opened 2022-11-30
release: v0.1.0

bhelx:


libextism.pc
```diff
@@ -5,6 +5,6 @@ libdir=${exec_prefix}/lib
 
 Name: extism
 Description: The Extism universal plug-in system.
-Version: 0.0.1
+Version: 0.1.0
```
bhelx, line 8: Is this right? Should this reflect libextism's version?

zshipko: Yeah I think that's right

PR #119, opened 2022-12-01
feat: Implement .NET Host SDK

mhmd-azeez:
I am working on a .NET host for extism. My plan is to do the following:
 - [x] Implement a proof of concept to make sure things are possible
 - [x] Write docs for the C# API so that the users get a nice IDE experience
 - [x] Create a github action to publish the NuGet packages
    - [x] Edit `ci.yml` to include .NET Sdk
    - [x] Create `release-dotnet.yml` to release `Extism.Sdk` nuget package
    - [x] Maybe Create `release-dotnet-native.yml` to release `Extism.runtime.win` nuget package
 - [x] Test on Linux (Help needed)
 - [x] Test on Mac (Help needed)
 - [x] Expose all of the Extism functions
 - [x] Write automated tests
 - [x] ~Edit README show that the there is a .NET SDK~. Probably we should not do this until we have a docs page.
 - [x] ~Use the `Extism.runtime.win-x64` package in the sample project~

Out of scope for this PR:
 - Json Serialization/Desererialization support


alistairjevans:
Hey @mhmd-azeez; external observer, thanks for raising this PR! Let me know if you want any assistance, I've got use-cases in mind...

Have you considered generating a native-assembly nuget package that contains a pre-built extism binary for various platforms? Similar to how (https://github.com/jedisct1/libsodium/actions/runs/3560505523)[libsodium does it]? That would give a nice out-of-the-box experience for .NET devs.  Your may choose not to include Android and iOS builds in there by default out of the pain of setting up those builds, but desktop and server platforms should be fairly straightforward?

mhmd-azeez:
Hello @alistairjevans I definitely plan to generate native assembly nuget packages. And if you're interested, we can work on this PR together.

@nilslice mentioned that they can drop the binaries in a folder in GitHub actions

alistairjevans:
Sure, I'm up for helping out; how you want to do it? I can fork your repo and merge into there to make this work? Or collaborate on your repo directly? Your choice.

mhmd-azeez:
Fantastic! @alistairjevans I just invited you to my repo, do you agree with the scope in the PR description?

alistairjevans:
Yeah, I think the scope makes sense; to my mind, step 1 is to build a solid (but thin) interop layer over the imported methods, and get that CI and tests in place. There's always going to be people who want to be able to reach directly for the methods that work with raw `ReadOnlySpan<byte>`.  That's more or less what you already have, and what this PR is for.

Once that's in place, a higher-level API that helps with serialising data to/from the plugins, a POCO representation of the manifest, and so on, will help most users pick this up more easily.  I haven't seen that there's a fixed standard for passing structured types to/from the PDK; JSON seems to be included in the PDK by default, but it's a relatively inefficient format for passing arguments, so I can imagine guidance (codegen?) in future for defining interop data structures.

I wonder if @nilslice has considered anything like a protobuf-style spec for defining parameters to plugin methods; this would allow you to arrive at a standard structure definition for all the supported PDKs from a single source.

The use-case I'm thinking about for this is plugins for processing network packets in near-real-time, so would want to keep the method passing overheads.

It looks like you can only pass a single data pointer to each plugin method, and I'm assuming that the data is copied into the WebAssembly context rather than providing direct access to the provided buffer.

bhelx:
Hey @mhmd-azeez I'd be happy to take some of these tasks if you don't want to do them:

* Test on Linux (Help needed)
* Test on Mac (Help needed)
* Create a github action to publish the NuGet packages (we can share this one or I can do it for you. I'll get an account and secret setup at least)

bhelx:
Okay we now have a Nuget API key stored as github secret: `NUGET_API_KEY`
Should be available from actions.

mhmd-azeez:
@bhelx Thank you very much! Testing is something your help will speed things up and as for the pipeline I'll need your help in fetching the native binaries. Will let you know once I get there

bhelx:
> > I'll need your help in fetching the native binaries. Will let you know once I get there

I can look into this as well. we may have to build it on the fly

mhmd-azeez:
@bhelx how can I make sure that logging works? currently I set log level to `Trace` and the `extism_log_file` returns true, but the log file is always empty, even when I pass in an incomplete wasm byte array

bhelx:
> Edit: I just pushed some code that I used to test the logging, I haven't been able to force it to write anything to the log file so far.

I seem to recall @nilslice or @zshipko may have something to say on this. I noticed this issue in another host library. May be a bug that we should track and fix. If it's returning true though I say keep going and don't worry about it! It should error out if you are doing it wrong

zshipko:
Looks like the logging issue is a bug - I just made a PR with a fix: https://github.com/extism/extism/pull/123

mhmd-azeez:
For the nuget packages, I think [OpenCVSharp](https://github.com/shimat/opencvsharp)'s approach is a good fit because Extism's binaries are pretty big (10 - 30 MB per platform). So we shouldn't force people to copy all of these binaries with their apps. Here is what I have in mind:

1. ExtismSharp: Contains the .NET host, targets .NET Standard 2.1. Later on if there is demand we can lower it to .NET Standard 2.0. But for now, I want to support `Span` and don't want to litter the codebase with `#if`s.
2. ExtismSharp.runtime.win: Contains the win-x64 binary. @bhelx @nilslice Do you think we should include the MSVC binary or the gnu one? Or should we offer both choices?
3. ExtismSharp.runtime.linux: Contains one of the linux binaries. @bhelx @nilslice any suggestions on which one to include?
4. ExtismSharp.runtime.macOS: Contains both `aarch64` and `x86_64` apple binaries.

@alistairjevans what do you think?

Note 1: I think if we go through that route, we should publish the runtime nuget packages in the https://github.com/extism/extism/actions/workflows/release.yml pipeline since these packages deployments don't have anything to do with the C# code. @bhelx @nilslice

Note 2: Including the binaries is mostly important for Windows, because extism already supports systemwide installation for macOS and linux. So we can even skip them. @bhelx @nilslice

nilslice:
> 2. `ExtismSharp.runtime.win`: Contains the win-x64 binary. @bhelx @nilslice Do you think we should include the MSVC binary or the gnu one? Or should we offer both choices?

I am not sure I really know the impact of deciding either one. Are there many Windows users who use the `gnu` toolchain vs. MSVC? Sorry for my naivety on the subject :) 

> Note 2: Including the binaries is mostly important for Windows, because extism already supports systemwide installation for macOS and linux. So we can even skip them. @bhelx @nilslice

I think we can skip these for macos and linux. This is assuming that dotnet doesn't expect any special treatment on those platforms with regard to how it locates shared libraries. I think it would use the system's linker anyway?

alistairjevans:
Do you think there's a reason not to lay claim to the `Extism` prefix rather than `ExtismSharp`? There's unlikely to be a "non-sharp" nuget package separate to this one.

I would then be tempted to ship `Extism.Host` for the host library, to give you space for `Extism.Pdk`.

I think storing the runtime packages separately is a good idea.  Would suggest a couple of things:

- The win package should probably include an arm64 build of extism as well if possible, there are a selection of Windows arm64 devices out there.  Will extism build for Windows arm64? Since it's rust, should be fairly straightforward?

- Using the MSVC compiler may give better windows support, and is available out-of-the-box on GitHub runners I believe. Documentation will need to indicate the VC++ redist package is needed though.

- Would propose an additional, optional, `Extism.runtime.all` which pulls in all supported runtime packages for those unlucky few (like myself), that deploy an app onto all of the above. It allows updates of the package to pull in any new supported platforms without needing to explicitly add them. Useful for devs building apps for user devices.

- I'd personally say that it would be good to keep that Linux runtime package, specifically to support the single-file publishing model; makes it easy to distribute apps to client Linux devices that way. Are we expecting a generic 'linux' package to be sufficient? Or are we going to need distro-specific ones?

nilslice:
+1 to naming `Extism` over `ExtismSharp`, but if this is a pattern that is more welcome and familiar to dotnet ecosystem, I don't mind it.

mhmd-azeez:
@nilslice @alistairjevans 

> I'd personally say that it would be good to keep that Linux runtime package, specifically to support the single-file publishing model; makes it easy to distribute apps to client Linux devices that way. Are we expecting a generic 'linux' package to be sufficient? Or are we going to need distro-specific ones?

Since publishing these packages is fairly straightforward, I think we should just publish for all 3 platforms.

> The win package should probably include an arm64 build of extism as well if possible, there are a selection of Windows arm64 devices out there. Will extism build for Windows arm64? Since it's rust, should be fairly straightforward?

I didn't include it because release artifacts don't have it. If @nilslice @bhelx think it's possible, then we should definitely add it.

@alistairjevans I agree with your naming suggestions, I will rename the projects accordingly

nilslice:
> I would then be tempted to ship Extism.Host for the host library, to give you space for Extism.Pdk.

I don't have strong feelings against using the `Host` name here, but we typically have referred to these as "SDKs" in other places.`Extism.Sdk` and `Extism.Pdk` (when we have one :)) may align better in the bigger picture. If folks think `Host` is more clear though I'm on board! 

> Will extism build for Windows arm64? 

Rust can target that platform, but I don't see a release for it for one of our dependencies `wasmtime`:

https://github.com/bytecodealliance/wasmtime/releases/tag/v3.0.1

I haven't tested to see if it wouldn't build, but I'd imagine `wasmtime` would release a build for it if it could target that platform too.

mhmd-azeez:
> I don't have strong feelings against using the Host name here, but we typically have referred to these as "SDKs" in other places.Extism.Sdk and Extism.Pdk (when we have one :)) may align better in the bigger picture. If folks think Host is more clear though I'm on board!

My vote is for `Extism.Host`

bhelx:
> Contains the win-x64 binary. @bhelx @nilslice Do you think we should include the MSVC binary or the gnu one? Or should we offer both choices?

Sorry catching up here. I understand the reason we may want to put the dll in the package for windows users, since they expect that. but for everything else we should probably follow the pattern with the rest of the libraries. Let the FFI library search out the dylib or so on the load path. Ideally we don't have to build for every potential architecture to and OS just to ship out the csharp library. It could get very big and we won't easily be able to future proof it.

bhelx:
> On second thought, maybe we should keep things consistent with the other platforms, @alistairjevans @bhelx what do you think?

Yeah I think consistency is key unless there is a strong idiomatic (dotnet) reason to break it.

bhelx:
> Rust can target that platform, but I don't see a release for it for one of our dependencies wasmtime:

man one other thing i need to do, make a table of architectures and OSs we support.

alistairjevans:
Happy with that argument @bhelx. Extism.Sdk it is then.

Re: distributing the binaries; on non-windows system, what's the deployment approach? Build from source on-device and put the built binary in the load path?

I guess to keep things simple and consistent we can start with that, but dotnet users are generally less familiar with build-from-source-libraries on any platform, so I can see some challenges on that front in the future.

bhelx:
> Re: distributing the binaries; on non-windows system, what's the deployment approach? Build from source on-device and put the built binary in the load path?

For linux and mac, we have an installer that manages the dynamic library. it puts it on your load path: https://extism.org/docs/install#macos-1

The SDKs find this library at runtime. For windows I'd defer to your judgement. We probably want to have some code that just loads the packaged version for windows and finds the installed version for other platforms.

mhmd-azeez:
> Ideally we don't have to build for every potential architecture to and OS just to ship out the csharp library. It could get very big and we won't easily be able to future proof it.

I don't mind going with the approach of the windows runtime nuget package. But I just wanted to clarify the process a bit so that both @bhelx and @nilslice have a clearer understanding of the future maintenance work.

So .NET has its own search path in addition to the OS search path, it's something like this: `runtimes\[runtime identifier]\native\ [native library file]`. So the nuget runtime packages will have a `runtime` folder that matches the search path. For example:

The `Extism.runtime.win` package will have this structure:

```
runtimes/
├─ win-x64/
│  ├─ native/
│  │  ├─ extism.dll
```

The `Extism.runtime.osx` package will have this structure:
```
runtimes/
├─ osx-x64/
│  ├─ native/
│  │  ├─ extism.dylib
├─ osx-arm64/
│  ├─ native/
│  │  ├─ extism.dylib
```

The `Extism.runtime.linux` package will have this structure:

```
runtimes/
├─ linux-x64/
│  ├─ native/
│  │  ├─ extism.so
```

Now, we can either add all of these to a single nuget package, or separate them into different nuget pacakages. The important part is, as you might have noticed, these nuget packages contain nothing .NET specific, so they can be published whenever a new version of `extism` is released, i.e. in the `Release` or `ci` pipeline.

We will have two other nuget packages: `Extism.Sdk` and `Extism.Pdk` that contain the .NET specific code and the release cycles can be decoupled from the runtime packages.

Now, whichever path you decide is totally fine with me because I understand your concerns as the library maintainers.

mhmd-azeez:
> We probably want to have some code that just loads the packaged version for windows and finds the installed version for other platforms.

If I am not mistaken, we don't need to do anything special. .NET should be able to find the native library on its own on macOS and Linux

alistairjevans:
@mhmd-azeez, correct the DllImport code will search the dynamic load paths on Linux and macOS, so that should just work.

On Windows it's considered very bad form these days to put anything in the system32 directory, so the extism dll will have to sit adjacent to the built exe on platform-specific RIDs or in that described folder structure on runtime-dependent builds.

mhmd-azeez:
Okay, let's go with only releasing the windows runtime package for now, we can always add the others later on if necessary

bhelx:
Sounds like a plan for now. My primary concern is we get into a situation where we have to ship this library with the runtime for each potential target. It could add up to like 500MB.

alistairjevans:
I think @mhmd-azeez's package layout proposal would require only downloading the packages they needed, since their dotnet project would reference:

- Extism.Sdk
- Extism.runtime.linux

To only download the windows-built binary to use with the Sdk, when they build their project.

We can further subdivide the packages into architecture-specific ones if required, so users can reference:

- Extism.Sdk
- Extism.runtime.linux.arm64

And so on.

So there's no need to require all of them to be downloaded at once, we would just push nuget packages for each runtime we support.

mhmd-azeez:
@bhelx @nilslice can you please approve the workflow for this PR? I want to change the CI workflow to build & test the .NET SDK too

dotnet/src/Extism.Sdk/Context.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// </summary>
+    public IntPtr NativeHandle { get; }
+
+    /// <summary>
+    /// Loads an Extism <see cref="Plugin"/>.
+    /// </summary>
+    /// <param name="wasm">The plugin WASM bytes.</param>
+    /// <param name="withWasi">Enable/Disable WASI.</param>
+    /// <returns></returns>
+    public Plugin CreatePlugin(ReadOnlySpan<byte> wasm, bool withWasi)
```
bhelx, line 31: I think this is good for now, we'll want to eventually support passing in raw bytes or a manifest object: https://extism.org/docs/concepts/manifest/

extism_plugin_new takes a byte array that can be either the raw bytes of the module or a json serialized Manifest.

bhelx:
I'm unsure how to approve the workflow. Perhaps @nilslice knows. He may be headed to the airport though. I'll see if i can figure it out.

nilslice:
Yea I checked from my phone (am en route to the airport) and can't find the button to run it.

dotnet/src/Extism.Sdk/Context.cs
mhmd-azeez: I think I will just update the docs that both are supported, in another PR we can add another layer `Extism.Sdk.Json` that adds full Json Serialization/Deserialization support and makes the config a proper object too

bhelx:
I found it. approved and ran. If we need to do this everytime, maybe there is a better way to do this.

bhelx:
Looks alright to me https://github.com/extism/extism/actions/runs/3605298556

mhmd-azeez:
@bhelx I will run any other workflows on my fork, it seems to work

mhmd-azeez:
> Looks alright to me https://github.com/extism/extism/actions/runs/3605298556
Yup, it works on both macOS and Linux. For Windows, we will need the runtime package which I will work on some other time

nilslice:
I also found it on my phone and it seemed like there was one more workflow to run. It's also scoped only to first-time contributors:

> First-time contributors require maintainers to approve workflows in order to run.

mhmd-azeez:
@nilslice Yeah, please don't run it anymore. Running it on my fork is good enough for now

mhmd-azeez:
Hello @bhelx, I have created two workflows:
 - `release-dotnet`: Builds, tests, packs, and publishes the Extism.Sdk nuget package. This should work as expected
 - `release-dotnet-native`: Packs and publishes the Extism.runtime.win-x64 nuget package. However, one thing is missing. It expects a `win-x64.dll` which is the MSVC extism dll

bhelx:
Testing this out on M1 macOS. I was able to get the sample to work by running:

```
> dotnet run --project samples/Extism.Sdk.Sample/
Extism version 0.0.1-rc.6
Input: Hello World
{"count": 3}
```

I'm curious where it's getting the runtime from I don't think I have this older rc version on my machine. Can you explain how this links to the shared library on mac and linux?

mhmd-azeez:
@bhelx I am not sure, but I am assuming .NET's searches in the usual places for shared libraries depending on the OS plus some additional places, the additional places as far as I am aware are:

- the binary folder
- the binary folder/runtime/[rid]/native

That version should 99% exist on your machine somewhere since I am not including any native libraries with the SDK. This article has some info on the resolution, but doesn't include all the different search paths:
 https://learn.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform

bhelx:
Thanks for the article. Yes, turns out problem was on my machine. I seem to be getting a sigkill now though. It exits without printing anything:

```
[I] ben@m1lap ~/d/e/dotnet (feature/dotnet-host)> dotnet run --project samples/Extism.Sdk.Sample/
[I] ben@m1lap ~/d/e/dotnet (feature/dotnet-host) [SIGKILL]> 
```

Perhaps it's not finding the library or there is something wrong with this version. Are you testing with extism runtime v0.1.0?

mhmd-azeez:
@bhelx yeah, not sure why that's happening. .NET throws an exception is printed to stdout when it can't find a native library:

```bash
Mo in W:\oss\extism\dotnet\samples\Extism.Sdk.Sample on feature/dotnet-host ● λ dotnet run
Unhandled exception. System.DllNotFoundException: Unable to load DLL 'extism' or one of its dependencies: The specified module could not be found. (0x8007007E)
   at Extism.Sdk.Native.LibExtism.extism_version()
   at Extism.Sdk.Native.Context.GetExtismVersion() in W:\oss\extism\dotnet\src\Extism.Sdk\Context.cs:line 102
   at Program.<Main>$(String[] args) in W:\oss\extism\dotnet\samples\Extism.Sdk.Sample\Program.cs:line 6
   at Program.<Main>(String[] args)
```

bhelx:
Yes, that's why i assume its a panic. no error message.

bhelx:
If i remove the library:

```
sudo rm /usr/local/lib/libextism.dylib 
```

and run again, I get the same `Unable to load shared library 'extism' or one of its dependencies.` as above. So it appears to be loading the library then panicking. Since it seemed to work with 0.0.1-rc.6, my assumption is there is some kind of mismatch in the interface or there is some type or mishandled pointer issue.

Which version of the runtime are you testing with @mhmd-azeez ?

mhmd-azeez:
@bhelx I am using the gcc compiled v.0.1.0 version for windows:

```
Mo in W:\oss\extism\dotnet\samples\Extism.Sdk.Sample on feature/dotnet-host λ dotnet run
Extism version 0.1.0
Input: example  
{"count": 3}
``

**Comment by bhelx** [comment]
Alright, I got it working:

```
> dotnet run --project samples/Extism.Sdk.Sample/
Extism version 0.1.0
Input: asldkfjsadf
{"count": 2}
```

But honestly I'm not sure how 😆 

I'm gonna assume this is a problem with my machine and not your code though. Carry on! I'll do some testing here on the sample project.

mhmd-azeez:
@bhelx I think this PR is done, only thing remaining is the `release-dotnet-native.yml` workflow. It expects `dotnet/nuget/runtimes/win-x64.dll` which is the MSVC extism dll. You mentione that you can fetch it/compile during the github action. Can you take a look at it? I prefer if we fetch if we are already storing it in an artifact.

Beside that, I will try to write some notes about releasing the nuget packages and I also want to release a version of the `Extism.runtime.win-x64` package so that I can reference it in the nuget package. Right now Windows users have to manually copy the dll to their bin folder which is defenitely something that .NET devs are not used to

mhmd-azeez:
@bhelx @nilslice I have added both of you to my fork

bhelx:
My availability will be spotty until Monday. I just did a little messing around with the library on Mac and didn’t find any issues. I can do more testing and write tests on Monday, but I think it’s probably in an good spot for beta. Perhaps Steve can take a look? Either way I’d say let’s get ready to get this merged Monday or early next week. Whenever you are ready. Amazing work!

bhelx:
Also I agree, probably best to fetch the DLL from latest release rather than build on the fly. What do you think @nilslice?

nilslice:
> Also I agree, probably best to fetch the DLL from latest release rather than build on the fly.

It will certainly be faster - and I _think_ we always want to ship an SDK to target the latest release anyway. The only tricky thing I can think of is when we make a runtime release, we have to also re-release the .NET SDK so that it bundles the latest runtime. This might turn out to be true for most runtime releases anyways (especially for runtime feature additions/changes) but worth calling out and thinking through now.  

Awesome work @mhmd-azeez!!

mhmd-azeez:
@bhelx @nilslice thanks!

> The only tricky thing I can think of is when we make a runtime release, we have to also re-release the .NET SDK so that it bundles the latest runtime

Just wanted to clarify one thing, the .NET SDK doesn't bundle the runtime. The `Extism.Sdk` nuget package doesn't actually depend on `Extism.runtime.win-x64` at all and can be used separately. The `Extism.runtime.win-x64` is just the runtime without any knowledge of the SDK. Whenever a project references `Extism.runtime.win-x64` it just copies the win-x64 dll to the app's output directory.

So the release cycle of `Extism.runtime.win-x64` is coupled with the release cycle of the native extism library. But that's not necessarily true for `Extism.Sdk`. I'll write some notes about both packages later on to make it clear when they should be published and what are the steps (mainly changing the version in the `.csproj`)

mhmd-azeez:
@bhelx @nilslice thanks!

> The only tricky thing I can think of is when we make a runtime release, we have to also re-release the .NET SDK so that it bundles the latest runtime

Just wanted to clarify one thing, the .NET SDK doesn't bundle the runtime. The `Extism.Sdk` nuget package doesn't actually depend on `Extism.runtime.win-x64` at all and can be used separately. The `Extism.runtime.win-x64` is just the runtime without any knowledge of the SDK. Whenever a project references `Extism.runtime.win-x64` it just copies the win-x64 dll to the apps output directory.

So the release cycle of `Extism.runtime.win-x64` is coupled with the release cycle of the native extism library. But that's not necessarily true for `Extism.Sdk`

alistairjevans:
I'll take a look on Monday @mhmd-azeez;  I'm away-from-keyboard until then I'm afraid.

nilslice:
@mhmd-azeez regarding: 
>  Test on more wasm plugins

I have some pre-compiled wasm modules in this repo, which I use as an example. 

https://github.com/extism/extism-fsnotify/tree/main/apps

Both plugins use these datatypes as input/ouput in JSON: 
https://github.com/extism/extism-fsnotify/blob/main/main.go#L17-L26

It may be useful to re-use them!

mhmd-azeez:
Hello @bhelx 

> @mhmd-azeez trying to understand what is needed from us to get this published. Are you saying this is how it works now? Or do you need us to work on the CI task to get the dll in the right place?

Unfortunately it doesn't work like that right now, so you'll need to work on the CI task to get the dll in the right place. If you can put the MSVC dll in `dotnet/nuget/runtimes/win-x64.dll`, everything should work as expected. If you can also publish a version for the runtime and SDK package it would be great so that I can reference them in the Sample projec. If not, we can do this in another PR. After that I think this PR is ready to be merged

mhmd-azeez:
After this PR is merged, I will try to work on adding a page on the extism docs about the .NET SDK

bhelx:
Hey @mhmd-azeez I have some notes on the interface: https://github.com/mhmd-azeez/extism/pull/3

In the rest of the libraries, we hide the implementation details of libextism. no need to expose length and error for example. I changed the interface of `call` to be bytes in and bytes out. The error code triggers an exception. This is more inline with how our other hosts do it. But if you think there is a more idiomatic way to do it let me know. Also feel free to merge and change it up or implement it yourself if you think there are better ways to accomplish this.

bhelx:
> After this PR is merged, I will try to work on adding a page on the extism docs about the .NET SDK

our docs page isn't pubic yet. I'm happy to do that for you. I can send a preview link here for you to approve.

bhelx:
Heads Up: Github does not seem to want to allow us to run these release workflows until we merge. So we can merge this when the code is ready and follow up to fix the workflows.

nilslice (dismissed):
Wow, great work on this!

mhmd-azeez:
> The branch with the doc changes is here: https://dotnet-sdk-doc.extism-docs.pages.dev/docs/integrate-into-your-codebase/dotnet-host-sdk/
> 
> Will update this after we merge, publish, and test it. Once we verify the instructions are working we'll go live with it.

@bhelx thank you very much. This makes me so happy :'D

A couple of notes:
1. Change `dotnet add package Extism.SDK --version 0.1.0` to `dotnet add package Extism.Sdk --prerelease`. The convention for package names is to be PascalCase and specifying `--prelease` allows people to install the package without us having to update the docs every time we push a new version.
2. I think we have to document that they will need the Extism library in the bin directory of the app on Windows or maybe #141 makes this better? Or after merging this PR and publishing the win runtime nuget package we can just ask them to add it to the project
3. Please update the code snippet to the one we have in the other branch so that it doesn't cause build errors

bhelx:
> After https://github.com/mhmd-azeez/extism/pull/3 is merged we should merge this PR too I thinkAfter https://github.com/mhmd-azeez/extism/pull/3 is merged we should merge this PR too I think

Yes, i think it's time to merge it and let's do some smaller, more focused, follow up PRs. Great work getting it this far! Ping me when you want to merge https://github.com/mhmd-azeez/extism/pull/3 . Is there anything else we need to do for it?

I'll update the docs site with your suggestions too, thank you!

bhelx:
I just realized i can merge it. Let's get all this merged up now! Just gonna wait on CI one more time.

bhelx:
Let's go!

mhmd-azeez:
Wohooo! 🎉🎉🎉

PR #122, opened 2022-12-02
feat(java-sdk): Create Java Host SDK

### LINKED ISSUE #117: Java SDK?
The README image suggests that Java (or maybe just Android?) is a "coming soon" target language. I'd be keen to know when that lands.

bhelx:
Closes #117


java/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.extism.sdk</groupId>
+  <artifactId>extism</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <name>extism</name>
+  <url>http://maven.apache.org</url>
+  <properties>
+    <maven.compiler.source>19</maven.compiler.source>
+    <maven.compiler.target>19</maven.compiler.target>
```
bhelx, line 12: I do not remember what this stuff means. I guess source can be recent. But what do I put for target? is it a minimum? Need to look it up

java/src/main/java/org/extism/sdk/LibExtism.java
```diff
@@ -0,0 +1,32 @@
+package org.extism.sdk;
+
+import com.sun.jna.Library;
+import com.sun.jna.Native;
+import com.sun.jna.Pointer;
+import com.sun.jna.ptr.IntByReference;
+//import com.sun.jna.Platform;
+
+public interface LibExtism extends Library {
+    LibExtism INSTANCE = (LibExtism) Native.loadLibrary("/usr/local/lib/libextism.dylib", LibExtism.class);
```
bhelx, line 10: There is probably a standard way to do this that is platform independent. Might be some code in JNA which does it or JNI. Will search it out.

java/src/main/java/org/extism/sdk/ExtismException.java
```diff
@@ -0,0 +1,15 @@
+package org.extism.sdk;
+
+public class ExtismException extends RuntimeException {
```
thomasdarimont, line 3: Need to add some javadoc here

java/src/main/java/org/extism/sdk/LibExtism.java
```diff
@@ -0,0 +1,135 @@
...
+    void extism_plugin_free(Pointer contextPointer, int pluginIndex);
+
+    /**
+     * Update plugin config values, this will merge with the existing values.
+     *
+     * @param contextPointer
+     * @param pluginIndex
+     * @param json
+     * @param jsonLength
+     * @return
```
thomasdarimont, line 132: what does the boolean return mean?

java/src/main/java/org/extism/sdk/LibExtism.java
```diff
@@ -0,0 +1,135 @@
...
+     * @return the result code of the plugin call.
+     */
+    int extism_plugin_call(Pointer contextPointer, int pluginIndex, String function_name, byte[] data, int dataLength);
+
+    /**
+     * Returns the length of a plugin's output data.
+     *
+     * @param contextPointer
+     * @param pluginIndex
+     * @return
```
thomasdarimont, line 90: .. return the length of the output data in bytes.

java/src/main/java/org/extism/sdk/manifest/Manifest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class Manifest {
+
+    private final List<WasmSource> wasm;
+
+    private final ManifestMemory memory;
```
thomasdarimont, line 14: Memory and allowedHosts is currently not used? We should either use this OR remove it for now.

thomasdarimont (changes requested):
Before finishing this, we could address the following points:

java/pom.xml
```diff
@@ -0,0 +1,82 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.extism.sdk</groupId>
+  <artifactId>extism</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
```
thomasdarimont, line 7: Should the version be aligned with the overall project version?

java/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+    </plugins>
+  </build>
+  <dependencies>
+    <dependency>
+      <groupId>net.java.dev.jna</groupId>
+      <artifactId>jna</artifactId>
+      <version>5.12.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
```
thomasdarimont, line 65: Do we actually need an JSON serialization library at all? If we only need to serialize the Manifest to JSON, perhaps a hand-crafted helper method will be sufficient.

java/src/main/java/org/extism/sdk/Context.java
```diff
@@ -0,0 +1,88 @@
...
+    public String getVersion() {
+        return LibExtism.INSTANCE.extism_version();
+    }
+
+    /**
+     * Get the error associated with a context, if plugin is {@literal null} then the context error will be returned.
+     * @param plugin
+     * @return
+     */
+    public String error(Plugin plugin) {
```
thomasdarimont, line 69: Perhaps an additional method like `public String error()` that explicitly returns the context error would be useful here.

bhelx:
@thomasdarimont in the past, i've published java libraries to OSSRH: https://central.sonatype.org/publish/publish-guide/#releasing-to-central

Is this still the best place to publish?

java/pom.xml
bhelx: No, the Host SDKs will be versioned independently based on semver.

java/src/main/java/org/extism/sdk/LibExtism.java
```diff
@@ -0,0 +1,135 @@
...
+     * Note: Memory for this plugin will be reset upon update.
+     *
+     * @param contextPointer
+     * @param pluginIndex
+     * @param wasm
+     * @param length
+     * @param withWASI
+     * @return
+     */
+    boolean extism_plugin_update(Pointer contextPointer, int pluginIndex, byte[] wasm, int length, boolean withWASI);
```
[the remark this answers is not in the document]
bhelx: same goes for what I wrote for `#call`

java/src/main/java/org/extism/sdk/support/JsonSerializer.java
```diff
@@ -0,0 +1,23 @@
...
+    static {
+        GSON = new GsonBuilder() //
+                .disableHtmlEscaping() //
+                .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES) //
+                .setPrettyPrinting() //
+                .create();
+    }
+
+    public static String toJson(Manifest manifest) {
+        return GSON.toJson(manifest);
```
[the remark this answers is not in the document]
bhelx: On the fence about it. I think eventually we will need to parse too. I'd be open to removing it but I can't say how long that will be feasible.

java/src/main/java/org/extism/sdk/wasm/WasmSource.java
```diff
@@ -0,0 +1,9 @@
+package org.extism.sdk.wasm;
+
+/**
+ * A named WASM source.
+ */
+public interface WasmSource {
+
+    String name();
```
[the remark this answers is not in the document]
bhelx: There is a json schema for the manifest and a page explaining the fields: https://extism.org/docs/concepts/manifest/

hash is optional.

thomasdarimont:
@bhelx Yes, releasing to maven central is still current, btw. the process of publishing releases can be simplified a lot by using a tool like [jreleaser](https://jreleaser.org) which makes [publishing libraries to maven central a breeze](https://jreleaser.org/guide/latest/examples/maven/maven-central.html).

java/pom.xml
thomasdarimont: Okay, then I'd recommend to follow a semver approach and use 3 component version number.

java/src/main/java/org/extism/sdk/wasm/WasmSource.java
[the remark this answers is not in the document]
thomasdarimont: Okay, I'll take a look.

java/src/main/java/org/extism/sdk/support/JsonSerializer.java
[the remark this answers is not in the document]
thomasdarimont: Quick idea:

1) Introduce a `JsonConverter` interface that can be used whenever we need to read / write json data. An instance of `JsonConverter` could be made available to the `Plugin` via the `Context`.

```java
interface JsonConverter {

    <T> String toJson(T object);
    <T> T fromJson(String json, Class<T> type);
}
```
2) Keep the gson library as dependency, but allow it to be excluded via dependency-excludes on the consumer side.

3) Implement a `GsonJsonConverter` implementation based on `JsonSerializer` 

4) Introduce a `ContextFactory` where a `JsonConverter` can be configured centrally with a `#setJsonConverter(..)` method. Then offer something like `ContextFactory#newContext()` to create new `Context` classes.

With this in place, if users prefer another JSON library and don't want to pull in gson, they could exclude it with a maven dependency exclude and simply configure their own `JsonConverter` implementation, e.g. based on 
fasterxml jackson.

Perhaps this looks a bit over-engineered, but I think having a `ContextFactory` to control creation of `Context` which could allow to share 
some additional information might be useful.

Alternative:
Keep the JsonSerializer class as is and only use it internally. If we need to read json data, just add another method like:

```
public <T> T fromJson(String json, Class<T> object) {
   return GSON.fromJson(json, object);
}
```

java/src/main/java/org/extism/sdk/LibExtism.java
```diff
@@ -0,0 +1,135 @@
...
+     * Calls a function from the @{@link Plugin} at the given {@code pluginIndex}.
+     *
+     * @param contextPointer
+     * @param pluginIndex
+     * @param function_name  is the function to call
+     * @param data           is the data input data
+     * @param dataLength     is the data input data length
+     * @return the result code of the plugin call.
+     */
+    int extism_plugin_call(Pointer contextPointer, int pluginIndex, String function_name, byte[] data, int dataLength);
```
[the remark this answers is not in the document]
thomasdarimont: That makes sense, what errors could happen here, appart from invalid native call parameterization?
Perhaps it would be helpful to have a few examples that could be used for unit tests.

thomasdarimont:
I think it would also make sense to use a common source code formatter to format the code properly, otherwise a slightly diffenent IDE setup will cause formatting noise in pull requests.

Perhaps a tool like [maven-checkstyle](https://maven.apache.org/plugins/maven-checkstyle-plugin/) could help here.
A checkstyle example config with sentive values can be found [here](https://github.com/adorsys/keycloak-config-cli/blob/main/checkstyle.xml) and [here (maven)](https://github.com/adorsys/keycloak-config-cli/blob/main/pom.xml#L367).

java/src/main/java/org/extism/sdk/Context.java
thomasdarimont: Makes sense, then protected is fine, the error can be used as the message of `ExtismException`.

java/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        <version>3.10.1</version>
+        <configuration>
+            <release>${java.version}</release>
+        </configuration>
+      </plugin>
+      <plugin>
+          <artifactId>maven-assembly-plugin</artifactId>
+          <configuration>
+              <descriptorRefs>
+                  <descriptorRef>jar-with-dependencies</descriptorRef>
```
thomasdarimont, line 28: It's usually recommended to let libraries pull in their dependencies explicitly, instead of creating a jar-with-dependencies.

java/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.10.1</version>
+        <configuration>
+            <release>${java.version}</release>
+        </configuration>
+      </plugin>
+      <plugin>
+          <artifactId>maven-assembly-plugin</artifactId>
```
thomasdarimont, line 25: why is the maven-assembly-plugin needed at all? The jar builds fine without it.

java/pom.xml
```diff
@@ -0,0 +1,82 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.extism.sdk</groupId>
+  <artifactId>extism</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <name>extism</name>
+  <url>https://github.com/extism/extism</url>
```
thomasdarimont, line 9: The project is lacking some metadata (organization, licensing, description, developer information) etc. that I think is necessary for releasing to maven central.

See here for a more [complete library with complete maven metadata](https://github.com/adorsys/keycloak-config-cli/blob/main/pom.xml#L18)

thomasdarimont:
For additional tests I need a few more example wasm files:
- an example for reading a value from the manifest config and returning that as a result
- a function that fails with an internal error

Perhaps you could create a generic "test" wasm file with multiple functions that could be used by all modules for consistent testing instead of having `code.wasm` in every sub-project.

thomasdarimont:
It seems as the ci fails due to unrelated issues in the Python SDK, perhaps a rebase with main would solve that.

thomasdarimont:
Just added another PR: https://github.com/extism/extism/pull/150
I think the libray is now in good shape, however I think it would be good to address the //FIXME comments (in the non-test) code before releasing this.

thomasdarimont:
Btw. the current state of the java-sdk requires, that users manually obtain the matching native extism library for their platform and store the library in a folder. This folder then needs to provided to the JVM via a System property e.g.  `-Djna.library.path=/path/to/folder`. or the lib must be located in the OS specific library search paths.

This should be documented in the readme.md of the SDK.

Alternatively, we could package the matching library for various platforms directly in the jar and use a tool like [native-lib-loader](https://github.com/scijava/native-lib-loader) to load the proper library for the given OS dynamically.

java/src/main/java/org/extism/sdk/support/JsonSerializer.java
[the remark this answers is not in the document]
bhelx: I think that's a good idea. I know how complex dependency clashing can get in Java. But I think it's fine to wait until after the first release. It should not affect the interface.

bhelx:
> Alternatively, we could package the matching library for various platforms directly in the jar and use a tool like [native-lib-loader](https://github.com/scijava/native-lib-loader) to load the proper library for the given OS dynamically. Would it be possible to provide the extism libraries for multiple os/platforms before the maven build runs?

Perhaps we could bundle the runtime natively for some people, but we'd prefer not to. You get an explosion of complexity doing that. At least for the time being, I think it's best for people to understand that they need this library installed somewhere where JNA can find it. We can offer them a property to override that if they get stuck.

aalmiray:
Hello there, I'm the author of JReleaser. The tool was originally conceived to be tightly coupled with a Git releaser hence why one must be configured. Disabling the sole releaser is the same as not having a releaser at all. However, one can have a releaser configured yet skip the release process.

Why do you need a releaser then? The tool uses the releaser information to calculate URLs such as issue tracker, releases, and other bits of information.

You may want to set the `skipRelease` property inside `<jreleaser><release><github>`. See https://jreleaser.org/guide/latest/configuration/release/github.html for more information. Additionally, the `prepare` step is used create files required by package managers. Maciej's guide shows the use of the `deploy` step instead. This `prepare` step is not the same as maven-release-plugin's `prepare`.

java/readme.md
```diff
@@ -0,0 +1,39 @@
...
+---
+
+Java SDK for the [extism](https://extism.org/) WebAssembly Plugin-System.
+
+# Build
+
+To build the extism java-sdk run the following command:
+
+```
+mvn clean verify
```
aalmiray, line 11: Good grief (@thomasdarimont  knows why 😉)

java/pom.xml
```diff
@@ -0,0 +1,188 @@
...
+            </plugin>
+            <plugin>
+              <groupId>org.jreleaser</groupId>
+              <artifactId>jreleaser-maven-plugin</artifactId>
+              <version>1.3.1</version>
+              <configuration>
+                <jreleaser>
+                  <release>
+                    <github>
+                      <enabled>false</enabled>
```
aalmiray, line 107: set `<skipRelease>true</skipRelease>` instead to keep the configured releaser and skip the creation of a Git release during the invocation of `jreleaser:release` or `jreleaser:full-release`. Invoking `jreleaser:deploy` only publishes JARs & POMs to Maven Central.

bhelx:
@aalmiray thanks for taking a look!

> Why do you need a releaser then? The tool uses the releaser information to calculate URLs such as issue tracker, releases, and other bits of information.

Perhaps we don't right this moment. The conflict right now is this is a mono-repo. The release channels are for the runtime. We will likely move this java repo out to it's own github repo one day.

aalmiray:
Given that you only want to deploy JARs and POMs then the default releaser as inferred by the environment should work. Just be sure to use the `skipRelease` property and you should be good to go. One more thing, the current settings for the deploy step are set such that auto close and release are disabled. You must log into the Sonatype UI and manually close and release the staged repository.

java/pom.xml
bhelx: This seemed to work for me. Thanks for the tip. FWIW i still had to set a token to a dummy value:

```
t extism: Execution default-cli of goal org.jreleaser:jreleaser-maven-plugin:1.3.1:deploy failed: JReleaser has not been properly configured.
[ERROR] github.token must not be blank. Configure a value using the Maven DSL, or define a JRELEASER_GITHUB_TOKEN environment variable, or define a key/value pair in /Users/ben/.jreleaser/config.yml with key = JRELEASER_GITHUB_TOKEN
```

aalmiray: Yes, that's true. A token is still required but as you noted it could be a any non-empty value. You'd need a true Github Personal Access Token when interacting with GitHub for real.

thomasdarimont:
Thanks @aalmiray for your help! I already created a small example project that shows the [java-sdk usage](https://github.com/thomasdarimont/extism-java-example).

thomasdarimont (commented):
I'd run `mvn verify` 
clean isn't necessary since it is a fresh checkout, verify runs any checks on results of integration tests to ensure quality criteria are met (if configured).

thomasdarimont:
@bhelx 

`java-version: '11'` in java section of ci.yaml should  be '17'.

aalmiray:
> @bhelx
> 
> `java-version: '11'` in java section of ci.yaml should be '17'.

It depends. What's the baseline for Java compatibility? If source/target/release compiler settings are not explicitly set then the baseline becomes the same as the JDK used to run the build. I'd assume the baseline should be as low as possible, 8 if the code permits it, otherwise 11.

thomasdarimont:
@aalmiray the pom currently uses  `<java.version>17</java.version>` with `<release>${java.version}</release>` and we currently use records in the code. We could potentially adapt the code to be compatible with Java 11. 

I'd skip Java 8 compat, for now :)

nilslice (approved):

bhelx:
So this branch has been hanging our CI on the python jobs. We think it was just something with this branch and not the code. hoping that merging will fix the problem. If not we may revert

bhelx:
Wow it's going green now: https://github.com/extism/extism/pull/175
turns out this branch was just haunted.

Thanks @thomasdarimont ! I'm going to work on docs now and some final cleanup items and push this out tomorrow

PR #123, opened 2022-12-02
fix: update crate name to extism_runtime in logging registration function

zshipko:
This bug exists because we didn't change the crate name in the logging registration function, so all of our logs from `extism_runtime` were being filtered out.


nilslice:
great catch @zshipko! maybe we should do a grep across all repos for any similar possible matches for the same issue.

nilslice (approved):

PR #124, opened 2022-12-02
feat: Improve usability of manifest types

zshipko:
- Adds some helper functions for creating a manifest, mostly helpful for the Rust SDK
- Renames `ManifestWasm` to `Wasm`
- Renames `ManifestMemory` to `MemoryOptions`
- `ManifestWasm` and `ManifestMemory` have been marked as deprecated
- Adds an alias from `MemoryOptions::max_pages` to `MemoryOptions::max` in serde specification

### DIFF
--- a/extism.go
+++ b/extism.go
@@ -53,11 +53,11 @@ type WasmFile struct {
 }
 
 type WasmUrl struct {
-	Url    string            `json:"url"`
-	Hash   string            `json:"hash,omitempty"`
-	Header map[string]string `json:"header,omitempty"`
-	Name   string            `json:"name,omitempty"`
-	Method string            `json:"method,omitempty"`
+	Url     string            `json:"url"`
+	Hash    string            `json:"hash,omitempty"`
+	Headers map[string]string `json:"headers,omitempty"`
+	Name    string            `json:"name,omitempty"`
+	Method  string            `json:"method,omitempty"`
 }
 
 type Wasm interface{}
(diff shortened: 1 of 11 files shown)


nilslice:
nice - just dropping a note to remind us that we will also  need to update: 
https://extism.org/docs/concepts/manifest

and the graphic that @bhelx generated. ben, what did you use to make that?

bhelx:
I believe it was this: https://jlblcc.github.io/json-schema-viewer/

There is probably a way to render it in the docs.

bhelx:
Since we've committed to the API, can we do this in a backwards compatible way? or save it for 1.0?

nilslice:
If we're going to make some changes like this, there's another I'd like to consider in the HttpRequest type: `header` -> `headers`

zshipko:
This PR should be backward compatible, with the old names marked as deprecated - the only thing changing in the JSON schema is the type names and `max` is changed `memory_max` (but max still works)

@steve - I will add `headers` as an alias for `header` right now

zshipko:
I just realized the `max`/`max_pages` and `header`/`headers` changes weren't present in the json schema because they were only aliases so I swapped the name of the struct fields and the alias names.

nilslice:
Added a commit to include similar name changes throughout the SDKs. Figure we should be consistent. This breaks backward-compatibility though, so we should discuss.

nilslice (approved):
LGTM if you're ok with it @zshipko

PR #138, opened 2022-12-05
refactor(haskell):  cleanup haskell SDK to prepare for release

zshipko:
- Switches from `stack` to `cabal`
- Cleanup SDK code
- Adds release action (still waiting on Hackage upload approval)


runtime/src/plugin.rs
```diff
@@ -251,44 +251,75 @@ impl Plugin {
...
+        // Check for Haskell runtime initialization functions
         // Initialize Haskell runtime if `hs_init` and `hs_exit` are present,
         // by calling the `hs_init` export
         if let Some(init) = self.get_func("hs_init") {
-            if self.get_func("hs_exit").is_some() {
-                // as a final check, ensure that the `hs_init` signature meets the exepectation
+            if let Some(cleanup) = self.get_func("hs_exit") {
                 if init.typed::<(i32, i32), (), _>(&self.memory.store).is_err() {
-                    return Ok(());
+                    trace!("hs_init found, but has the incorrect type");
```
nilslice, line 261: maybe we could instead serialize and log the function type here vs. saying "incorrect" -- if it's in there it could be correct to someone, just not for Haskell runtime detection.

runtime/src/plugin.rs
```diff
@@ -259,7 +259,10 @@ impl Plugin {
         if let Some(init) = self.get_func("hs_init") {
             if let Some(cleanup) = self.get_func("hs_exit") {
                 if init.typed::<(i32, i32), (), _>(&self.memory.store).is_err() {
-                    trace!("hs_init found, but has the incorrect type");
+                    trace!(
+                        "hs_init found with type {:?}, which doesn't match the type exported by the Haskell runtime",
```
nilslice, line 263: I'll push up another message here -- sorry to nit. My hesitation is that this could mislead someone who has no reason to think about the Haskell runtime. 

I know it would be unlikely for a non-Haskell module to export both `hs_init` and `hs_exit`, and therefore meet BOTH these conditions... but in the case that it did happen, we would be adding unrelated information to a log.

IMO, we don't need to say anything about the Haskell runtime until after `return Some(Runtime::Haskell { init, cleanup });`, where we can be pretty sure that whomever is looking at the `trace` message should see something about Haskell.

nilslice (dismissed):
Ok, @zshipko if you're on board with the updated message in the trace log, and CI passes - this LGTM!

zshipko:
Cool, I have a few more small changes I’ll push in the morning then I’ll merge it!

PR #140, opened 2022-12-06
fix: use `Dir::open_ambient_dir` instead of `Dir::from_std_file`

zshipko:
`Dir::from_std_file` works on Linux, but does not work as expected on Windows

### DIFF
--- a/runtime/src/plugin.rs
+++ b/runtime/src/plugin.rs
@@ -35,6 +35,7 @@ pub struct Wasi {
 impl Internal {
     fn new(manifest: &Manifest, wasi: bool) -> Result<Self, Error> {
         let wasi = if wasi {
+            let auth = wasmtime_wasi::ambient_authority();
             let mut ctx = wasmtime_wasi::WasiCtxBuilder::new();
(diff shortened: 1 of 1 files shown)


nilslice:
`ambient` is a great name for this!

noun. _relating to the immediate surroundings of something._

I have logged #141 to track improvements to our Windows support. I wonder if this PR would be helpful for wasmtime to see? I didn't even consider different checking cross platform behavior on this. Maybe there's a way to more clearly point it out (even some extra docs?).

sunfishcode:
Yes; ambient is a great name, though to be sure, we didn't invent it; it's common in capability-based security circles 😀.

The change here looks good to me. That said, I'd also be interestested in learning how `from_std_file` misbehaves on Windows, if you happen to have more info.

nilslice:
> it's common in capability-based security circles 

Oh interestng, TIL!  

An issue reported on [Discord](https://discord.com/channels/1011124058408112148/1011124061100843020/1049509012284129423) and the discussion thereafter has some additional context. Thanks for taking a look.

nilslice (approved):

PR #143, opened 2022-12-06
fix: disallow all hosts if `allowed_hosts` is empty

zshipko:


nilslice:
Nice! Does this already support the `*` pattern to signify all hosts are allowed?

I will update docs accordingly once this is out.

nilslice (approved):

PR #147, opened 2022-12-07
(java-sdk) Next round of refactoring

thomasdarimont:
- Context#error is now protected and used to extract the error message for the ExtismException
- ByteArrayWasmSource is now correctly serialized (wasm data is now encoded base64)
- Renamed ManifestMemory to MemoryOptions to align with rust codebase
- Introduced WasmSourceResolver to load wasm source
- Added unit tests
- Revised Unit tests
- Polished javadoc
- Added necessary metadata for release to maven central
- Removed maven-assembly-plugin as we want to allow consumers to control the dependencies


thomasdarimont:
@bhelx here is the next round of refactoring, additional javadoc and tests.

java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
```diff
@@ -0,0 +1,11 @@
+package org.extism.sdk.manifest;
+
+public record MemoryOptions(
+
+        /**
+         * Max number of pages.
+         */
+        Integer max // Integer to allow `null` as max
```
nilslice, line 8: It's yet to be release, but we recently renamed this to `max_pages` for better clarity. 

We could do this now, and align the release with an upcoming Extism release (though `max` will still work as an alias), or we could release the Java SDK with this as `max` and follow up with another release shortly thereafter with the new name. 

@bhelx @zshipko, any thoughts?

zshipko: I’m open to either. Just depends on if we want to make a release as soon as possible or wait for the next runtime release.

java/pom.xml
```diff
@@ -1,82 +1,110 @@
...
+    </licenses>
+
+    <organization>
+        <name>Dylibso, Inc.</name>
+        <url>https://dylib.so</url>
+    </organization>
+
+    <developers>
+        <developer>
+            <name>Benjamin Eckel</name>
```
bhelx, line 26: If anything your name should go here @thomasdarimont 😆 

But we've been following the pattern of setting maintainers as:

name: `The Extism Authors`
email: `oss@extism.org`

java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
bhelx: Also open to either, but my opinion is I'd prefer to wait until we have a few breaking changes grouped up. Unless we have a rush to get it out.

nilslice: Ok, I agree -- we can ship it with this as-is and release an update when we are ready. Thanks!

PR #148, opened 2022-12-07
feat: Add ability to register host functions with the runtime from the Rust SDK

zshipko:
- Adds the ability to initialize plugins with additional functions created by the host (only supported by the Rust SDK right now) 
- Adds `PluginBuilder` to the Rust SDK to make it easier to configure all the options

## Hello, world

The following example uses `println` from the host to print to stdout from inside the plugin.

Host:
```rust
use extism::*;

fn main() -> Result<(), Error> {
    let context = Context::new();
    let manifest = Manifest::new([std::path::PathBuf::from("code.wasm")]);
    let say_hello = Function::new("say_hello", [ValType::I64], [], |caller, input, _out| {
        let internal = caller.data();
        let r = internal.memory().get_str(input[0].unwrap_i64() as usize)?;
        println!("Hello, {r}!");
        Ok(())
    });
    let mut plugin = PluginBuilder::new(manifest)
        .with_function(say_hello)
        .with_wasi(true)
        .build(&context)?;
    plugin.call("test", b"world").unwrap();
    Ok(())
}
```

Plugin:
```rust
use extism_pdk::*;

extern "C" {
    fn say_hello(offs: u64);
}

#[plugin_fn]
pub fn test(input: String) -> FnResult<()> {
    let memory = Memory::from_bytes(&input);
    unsafe {
        say_hello(memory.offset);
    }
    Ok(())
}

```

### DIFF
--- a/manifest/src/lib.rs
+++ b/manifest/src/lib.rs
@@ -166,9 +166,9 @@ pub struct Manifest {
 
 impl Manifest {
     /// Create a new manifest
-    pub fn new(wasm: impl Into<Vec<Wasm>>) -> Manifest {
+    pub fn new(wasm: impl IntoIterator<Item = impl Into<Wasm>>) -> Manifest {
         Manifest {
-            wasm: wasm.into(),
+            wasm: wasm.into_iter().map(|x| x.into()).collect(),
             ..Default::default()
         }
     }
--- a/runtime/src/context.rs
+++ b/runtime/src/context.rs
@@ -16,53 +16,6 @@ pub struct Context {
 
 const START_REUSING_IDS: usize = 25;
 
-pub struct Function(
-    pub(crate) String,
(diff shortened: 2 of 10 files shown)


nilslice:
Finally tested this and did my own functions / plugin calls - works great! Opens up so many cool options. 

One thing that we will need to take into consideration (really just docs & awareness) is for the Playground - which will error out if a custom import is expected from a plug-in. I don't think there's really anything more we can do, bedsides maybe list those imports in the error and suggest that this plugin was made to use a custom host function. 

I want to play more with this before we release it, and add some e2e testing around it. But it LGTM to merge 🤘

nilslice (approved):

zshipko:
Great! I was thinking we should just put out a very basic implementation and see what we can build around it before committing to something more complex. So I agree, let’s merge this and make additional PRs for any improvements.

PR #150, opened 2022-12-08
(java-sdk) Add tests for Manifest and polish MemoryOptions

thomasdarimont:
- Fix broken javadoc comment
- Add tests for Manifest serilization and refactor tests
- Remove duplicate developer entry in pom.xml
- Add initial readme

Improves #117

### DIFF
--- a/java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
+++ b/java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
@@ -6,7 +6,7 @@ import com.google.gson.annotations.SerializedName;
  * Configures memory for the Wasm runtime.
  * Memory is described in units of pages (64KB) and represent contiguous chunks of addressable memory.
  *
- * @param maxPages Max number of pages.
+ * @param max Max number of pages.
  */
-public record MemoryOptions(@SerializedName("max_pages") Integer maxPages) {
+public record MemoryOptions(@SerializedName("max") Integer max) {
 }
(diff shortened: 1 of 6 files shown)


java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
```diff
@@ -1,11 +1,12 @@
...
-        Integer max // Integer to allow `null` as max
-//
-) {
+/**
+ * Configures memory for the Wasm runtime.
+ * Memory is described in units of pages (64KB) and represent contiguous chunks of addressable memory.
+ *
+ * @param maxPages Max number of pages.
+ */
+public record MemoryOptions(@SerializedName("max_pages") Integer maxPages) {
```
bhelx, line 11: @nilslice we decided to hold off on this change right? Just making sure.

nilslice: Yea, I think we will ship this as-is so it's immediately usable with the current release, and then we can make SDK-wide changes for the next release.

nilslice: by as-is, I mean using `"max"` as the name :)

bhelx: @thomasdarimont if you don't mind let's revert the change back to max for the time being.

thomasdarimont: I renamed the `MemoryOptions` field back to `max`.

PR #153, opened 2022-12-09
feat: Add `HttpRequest::timeout_ms` and `Manifest::http_timeout_ms` 

### LINKED ISSUE #151: Configurable HTTP timeouts
For `extism_http_request` in particular, we should add both host-configurable HTTP timeouts for call-outs made from a plug-in, as well as a way to set it from the PDK, but not to exceed one if set by the host. The PDK should only be able to set a timeout that is shorter than one set by the host.

zshipko:
Closes #151 

- Adds ability to configure timeout for `extism_http_request`
- Sets default timeout to 30s, but this can be overridden still
- Updates SDKs to support the new fields

### DIFF
--- a/runtime/src/manifest.rs
+++ b/runtime/src/manifest.rs
@@ -1,6 +1,7 @@
 use std::collections::BTreeMap;
 use std::fmt::Write as FmtWrite;
 use std::io::Read;
+use std::str::FromStr;
 
 use sha2::Digest;
 
@@ -137,18 +138,23 @@ fn to_module(engine: &Engine, wasm: &extism_manifest::Wasm) -> Result<(String, M
             #[cfg(feature = "register-http")]
             {
                 // Setup request
-                let mut req = ureq::request(method.as_deref().unwrap_or("GET"), url);
+                let method =
+                    reqwest::Method::from_bytes(method.as_deref().unwrap_or("GET").as_bytes())?;
+                let url = url::Url::parse(url)?;
+                let mut req = reqwest::blocking::Request::new(method, url);
 
                 for (k, v) in headers.iter() {
-                    req = req.set(k, v);
+                    let k = reqwest::header::HeaderName::from_str(k)?;
+                    let v = reqwest::header::HeaderValue::from_str(v)?;
+                    req.headers_mut().insert(k, v);
                 }
 
-                req = req.timeout(std::time::Duration::from_millis(
+                req.timeout_mut().insert(std::time::Duration::from_millis(
                     timeout_ms.unwrap_or(30000),
                 ));
 
                 // Fetch WASM code
-                let mut r = req.call()?.into_reader();
+                let mut r = reqwest::blocking::Client::new().execute(req)?;
                 let mut data = Vec::new();
                 r.read_to_end(&mut data)?;
 
--- a/runtime/src/manifest.rs
+++ b/runtime/src/manifest.rs
@@ -143,7 +143,7 @@ fn to_module(engine: &Engine, wasm: &extism_manifest::Wasm) -> Result<(String, M
                     req = req.set(k, v);
                 }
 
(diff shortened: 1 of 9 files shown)


runtime/src/pdk.rs
```diff
@@ -338,6 +338,19 @@ pub(crate) fn http_request(
             r = r.set(k, v);
         }
 
+        // Configure timeout
+        let global_timeout = data.plugin().manifest.as_ref().http_timeout_ms;
+        let req_timeout = req.timeout_ms;
+
+        r = r.timeout(std::time::Duration::from_micros(
```
nilslice, line 345: is `from_micros` here intentional?

runtime/src/manifest.rs
```diff
@@ -142,6 +143,10 @@ fn to_module(engine: &Engine, wasm: &extism_manifest::Wasm) -> Result<(String, M
                     req = req.set(k, v);
                 }
 
+                req = req.timeout(std::time::Duration::from_micros(
```
nilslice, line 146: also here, `from_micros` -- should this be `from_millis` ?

runtime/src/pdk.rs
zshipko: No, thanks for catching that!

nilslice: pushed up a fix!

nilslice:
Testing this out now, all is working great - just curious about a couple things: 

1. for a host that wants to override the default to effectively disable a timeout, letting a plugin wait for a response indefinitely, I believe the only way to do that would be to use [`Duration::MAX`](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.MAX)? I think that's reasonable.. and we can document that. Let me know if that's untrue though!
2. it seems as though there is no way for a plugin to recover from a host-imposed timeout. I can see how this would be useful as a control mechanism from the host, but think that we should instead lift that responsibility to a more general plugin timeout, and not cause an HTTP timeout to bail the whole plugin. 

re: 2, I'm looking at the implementation though and it seems like the `http_request` PDK function should just return the error to the caller within the plugin, but from my host, I can't get anything back except an err:

host:
```rust
use extism::*;

fn main() -> Result<(), Error> {
    let context = Context::new();
    let manifest = Manifest::new([std::path::PathBuf::from("code.wasm")])
        .with_http_timeout(std::time::Duration::from_millis(1000));

    let mut plugin = PluginBuilder::new(manifest).build(&context)?;

    match plugin.call("test", &[]) {
        Err(e) => println!("err: {:?}", e),
        Ok(s) => println!("output: {}", String::from_utf8(s.to_vec()).unwrap()),
    }

    return Ok(());
}
```

plugin: 

```rust
use extism_pdk::*;

#[plugin_fn]
pub fn test(_input: ()) -> FnResult<String> {
    let req = HttpRequest::new("https://httpstat.us/200?sleep=2000");
    match http::request::<()>(&req, None) {
        Err(_) => Ok(format!("error!! but sending Result::Ok")),
        Ok(res) => Ok(format!("status: {}", res.status_code())),
    }
}
```

**Note**: both match arms in the plugin return `Ok`, so the runtime is propagating the error up before the plugin can do anything about it.

zshipko:
I think setting timeout to 0 will also wait indefinitely, I thought I saw that in the docs but can’t find it right now.

I can test that out and look into the error propagation issue tomorrow - I have a pretty good idea of how to fix that and I will update this PR.

nilslice:
Ok sounds good! Thanks. I'll also take a look to see about the 0 timeout setting in the http client.

nilslice:
update: setting the timeout to `0` bails immediately with the timeout error

zshipko:
Okay, thanks for checking! Looks like `Duration::MAX` is the best option then.

bhelx:
We're closing this now that we have a top level function timeout. If anyone does feel we need to support this, we can re-open.

PR #154, opened 2022-12-11
fix: remove old/duplicate Haskell runtime initialization code

zshipko:


nilslice (commented):
@bhelx wasnt there some mono / c# runtime code that needs to be started upon plugin initialization? wondering if this is applicable there.

zshipko:
We should use the new `Runtime` enum for that. This is code that was supposed to be removed after adding the nicer runtime initialization interface.

nilslice:
Definitely agree - mostly leaving this as a note to connect some dots between here and convos on Discord. 

But I also didn't realize this code was so specific to Haskell on first glance 😆

nilslice (approved):

bhelx:
Yeah, the Csharp PDK will be based on mono. It needs `_start` to be called to initialize everything or else it crashes. I'm a bit confused why the runtime doesn't do it when you enable wasi. Will follow back up when we get further along with development. It could be a quirk with that one PDK.

zshipko:
I don't think WASI will really initialize language specific runtimes - to initialize Haskell `hs_init` is called in `_start`, I was able to re-export that function by passing `--export=hs_init` to `wasm-ld` during compilation. Then calling `hs_init` lets us call other functions without going through `_start` - maybe there is something similar for mono?

PR #163, opened 2022-12-13
feat: Add option to set timeout for plugin 

zshipko:
- Adds `Manifest::timeout_ms` field to specify the plugin timeout in milliseconds
- Sets a 30 second default timeout

### DIFF
--- a/manifest/schema.json
+++ b/manifest/schema.json
@@ -40,6 +40,14 @@
         }
       ]
     },
+    "timeout_ms": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "format": "uint64",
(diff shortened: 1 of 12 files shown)


nilslice (approved):

nilslice:
I think we can do a little more work on the context/manifest API - maybe just to add some more methods here and there for changing the timeout between plugin calls etc. (if possible).. but I think this is ready to merge if you do!

zshipko:
Sounds good! I will merge this now and we can figure out some improvements after testing it out.

One possible future improvement could be the ability to set a default timeout at the context level and setting a plugin timeout would override that.

PR #169, opened 2022-12-15
fix: use timeouts for language-specific runtime initialization/finalization calls

zshipko:
After adding the timer code plugins written in Haskell fail with a timeout when trying to call `hs_init` - this PR fixes the runtime initialization and updates those functions to respect the configured timeout.


nilslice (approved):
great catch!

PR #176, opened 2022-12-19
chore: move each language to separate workflow

zshipko:
- Moves each language into a separate workflow
- Only runs CI for languages when they're changed (or when the runtime is changed)


nilslice (approved):
:1st_place_medal:

PR #180, opened 2022-12-19
fix(elixir): use local Rust `extism` package from `extism_nif` for local development

zshipko:


elixir/native/extism_nif/Cargo.toml
```diff
@@ -11,5 +11,5 @@ crate-type = ["cdylib"]
 
 [dependencies]
 rustler = "0.26.0"
-extism = { version = "0.1.0" }
+extism = { version = "0.1.0", path = "../../../rust" }
```
bhelx, line 14: I had to remove this for a reason. Can't quite remember why. I think it was failing on release. I'll test it out tomorrow.

nilslice: @bhelx - did you get a chance to test this out? just don't want to let it sit too long

bhelx: if i remember correctly, it was breaking on release. let me see if there is a way i can dry run it

bhelx: `mix hex.publish --dry-run` worked

PR #186, opened 2022-12-27
feat(zig-sdk): Create Zig Host Sdk

### LINKED ISSUE #126: Zig SDK?
Congrats for Extism, which looks like a really useful project.

However, especially given the number of supported languages, and its maturity regarding WebAssembly support, I was surprised to see Zig missing from the list.

Is a Zig SDK planned soon?

usdogu:
Will close #126 once done.

# TODOs
- [x] Doing anything that requires manifest
- [x] CI support
- [x] Multithreaded tests
- [x] Support for gyro and zigmod


zig/examples/basic.zig
```diff
@@ -0,0 +1,42 @@
+const std = @import("std");
+const testing = std.testing;
+const sdk = @import("extism-sdk");
```
nilslice, line 3: I will defer to your Zig expertise here, but conventionally we are naming the SDK packages `extism`, and then to refer to the top-level namespace as `extism`. So:

```zig
const extism = @import("extism");
const Context = extism.Context;
const Plugin = extism.Plugin;
// ... and so on
```

If there are idioms in the Zig ecosystem that would suggest otherwise, I think we should change this to match other SDKs.

zig/src/context.zig
```diff
@@ -0,0 +1,33 @@
...
+const c = @import("ffi.zig");
+const utils = @import("utils.zig");
+const toCstr = utils.toCstr;
+
+pub const Context = struct {
+    mutex: std.Thread.Mutex,
+    ctx: *c.ExtismContext,
+
+    // We have to use this until ziglang/zig#2647 is resolved.
+    error_info: ?[]const u8,
```
nilslice, line 11: I read through the issue, thank you for adding this note. I _think_ I understand, but to clarify, this `error_info` field could eventually move into an error set union such as: 

```zig
const ExtismError = union(error) {
  Context: []const u8,
  Plugin: []const u8,
  // ... etc
}
```

ref: https://github.com/ziglang/zig/issues/2647

zig/examples/basic.zig
usdogu: I named it without looking other libraries, will push a commit to rename to extism

zig/src/context.zig
usdogu: I want to have a union like that too but AFAIK, having a field for reporting errors is the best option we have right now

nilslice: Oh, I only meant to ask that you confirm my basic understanding of the proposal you mentioned :) 

I see this is the way we must implement it for now. Thanks!

nilslice:
Thank you, @usdogu! I'm planning day for review and testing this tomorrow. This has been a really fun way to start learning Zig, I appreciate all of your work here. 

Once we are ready to merge, would you be interested in helping review some of the docs I am preparing for both the PDK and this SDK? Then I can announce the support more broadly and share it with the Zig ecosystem.

usdogu:
I'll be happy to review, thanks to you for creating this awesome project!

usdogu:
I'll be happy to get assisted by someone with mac experience to fix the macos CI. The latest commit fixes the pkg-config, but linux also works fine without it.

nilslice:
@usdogu I will take a look at it later today. I think we are trying to move away from using pkg-config on the SDKs. I'm not super familiar with it, but don't think it's really a macOS thing anyway, so we can probably just remove it. 

On *nix platforms, we're relying on library installations being in existing lookup paths (e.g. `/usr/local/lib` etc). Or that the path is set via environment variable at run time. 

I only added the pkg-config settings for use with the Go SDK after doing some research about shipping CGO packages, but I think it actually complicates things more than it's helping.

nilslice:
@usdogu - sorry for delay here. I've needed to focus on a couple other things the past few days and have not had as much time to review this as I thought I would! Will do so within the next couple of days (this weekend or early next week!).

usdogu:
No problem! Can you approve running the workflow just so I can look for other things if that fails too?

nilslice:
> Can you approve running the workflow just so I can look for other things if that fails too?

I didn't realize I would need to approve it for each commit here. Sorry about that. I will see what settings I can adjust to make this easier :)

nilslice:
Ok, you should be able to run workflows here without re-approval.

usdogu:
thanks!

nilslice:
@usdogu - thank you again for contributing this! I've reviewed (though with still a Zig beginner's eye) and tested in a new host program. I will merge this now and open other PRs if I have any changes. 

I'll send over a link here to the docs preview with both the PDK and SDK documentation for your feedback.

nilslice (approved):

nilslice:
@usdogu, for your review when you are able to take a look: 
SDK: https://zig-docs.extism-docs.pages.dev/docs/integrate-into-your-codebase/zig-host-sdk/
PDK: https://zig-docs.extism-docs.pages.dev/docs/write-a-plugin/zig-pdk

PR #195, opened 2023-01-03
feat: Add C API for host functions + support for C++, Python, Go, Node, OCaml

### LINKED ISSUE #20: reminder: consider inlining -l/-L paths vs. pkg-config in go sdk
https://github.com/extism/extism/blob/e12efbeadb58b0a1dcd8e7838cf9e51eb8f2152a/extism.go#L12

zshipko:
- New types:
  - `ExtismValType` - Enum of WebAssembly types
  - `ExtismValUnion` - A union of the possible WebAssembly types
  - `ExtismVal` - A struct with `ExtismValType` and `ExtismValUnion`
  - `ExtismFunction` - The host function wrapper type
  - `ExtismFunctionType` - The type of the host function callback
  - `ExtismCurrentPlugin` - Provides access to the currently running plugin from inside a host function

- New functions:
  - `extism_function_new` - Create a new `ExtismFunction`
  - `extism_function_free` - Free an `ExtismFunction`
  - `extism_current_plugin_memory`, `extism_current_plugin_memory_alloc`, `extism_current_plugin_memory_free`, `extism_current_plugin_memory_length`  - Manage plugin memory from inside a host functions

- Updated functions
  - `extism_plugin_new` and `extsim_plugin_update` - now accept two extra parameters for `ExtismFunction*` array and length of that array

## Notes

- Host functions take a user-data argument, which is owned by the resulting `ExtismFunction` and will be cleaned up when `extism_function_free` is called (if a cleanup function was passed in with the user data)
- Host functions in every SDK require working with `ExtismVal` arguments directly,  this is pretty low-level for what is kind of a high-level feature. We could work on adding some types to the SDKs that make working with pointers to plugin data more accessible, maybe something similar to how the Rust PDK handes input/output data.
- In each language the host functions more-or-less share a signature: `(CurrentPlugin plugin, Val inputs[], Val outputs[], userData)` 
  - C, C++, OCaml and Go take a single userData argument but Python and Node take a "rest" argument which allows passing any number of user-data values
- Go requires the host function to be exported:  https://github.com/extism/extism/blob/f9eb5ed8395218f08627622517e801491efb2a53/go/main.go#L13-L26
- Zig and Ruby should be relatively simple to add host functions to next but I haven't really looked into Elixir, .NET or Java yet.
- Also closes #20


zshipko:
Just pushed the change to get rid of `plugin_new_with_functions`/`plugin_update_with_functions` and add the parameters to `extism_plugin_new`/`extism_plugin_update` and updated all the SDKs

mhmd-azeez:
@nilslice I will try to add bindings to the .NET sdk after the PR is merged and the new binaries are released

nilslice:
@mhmd-azeez that would be awesome.  thanks!

PR #196, opened 2023-01-03
feat: WASI support for browser runtime

### LINKED ISSUE #160: WASI support in the browser runtime
We currently don't support WASI

Screenshot from the playground:

image

Putting this issue here to track the problem.

bhelx:
closes #160

Goal is not to expose all the WASI functionality, but to just write minimum code to allow PDK targets that require WASI (e.g. tinygo compiled plugins) to work in the browser. We can follow up with allowing the user more control over the WASI environment.

### DIFF
--- a/browser/src/plugin.ts
+++ b/browser/src/plugin.ts
@@ -82,6 +82,9 @@ export default class ExtismPlugin {
       env: environment
     };
     this.module = await WebAssembly.instantiate(this.moduleData, env);
+    if (this.module.instance.exports._start) {
+      wasi.start(this.module.instance);
+    }
     return this.module;
   }
 
--- a/browser/src/plugin.ts
+++ b/browser/src/plugin.ts
@@ -1,9 +1,7 @@
 import Allocator from './allocator';
 import { PluginConfig } from './manifest';
-//@ts-ignore
-import WASI from "@bjorn3/browser_wasi_shim/src/wasi";
-//@ts-ignore
-import { File } from "@bjorn3/browser_wasi_shim/src/fs_core";
+//@ts-ignore TODO add types to this library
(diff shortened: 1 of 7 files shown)


nilslice (approved):

nilslice:
Oh, these also may need to be executed if they exist: https://github.com/extism/extism/blob/e3a427b0cc35155c6ecd7c52f9b05ae11071745c/runtime/src/plugin.rs#L344

sunfishcode:
> Oh, these also may need to be executed if they exist:
> 
> https://github.com/extism/extism/blob/e3a427b0cc35155c6ecd7c52f9b05ae11071745c/runtime/src/plugin.rs#L344

Any program meant to be run in a WASI environment shouldn't be expecting additional initialization calls beyond _start.

zshipko:
> Any program meant to be run in a WASI environment shouldn't be expecting additional initialization calls beyond _start.

Those calls are being done instead of `_start`. We found that when compiling Haskell to WASM you're able to export multiple functions and initializing with `hs_init` allows for those functions to be called directly. Is side-stepping the `_start` function like that an issue?

browser/src/plugin.ts
```diff
@@ -82,6 +82,9 @@ export default class ExtismPlugin {
       env: environment
     };
     this.module = await WebAssembly.instantiate(this.moduleData, env);
+    if (this.module.instance.exports._start) {
+      wasi.start(this.module.instance);
```
zshipko, line 86: I don't think we want to call `_start` here, since that will actually run the plugin, it should be called in `Plugin.call` instead

bhelx: @zshipko so, you mean it should adopt the same methodology we use in the Rust runtime?

PR #197, opened 2023-01-04
feat(zig): import extism.h directly instead of hand-written bindings

zshipko:


nilslice (approved):
Nice - this will help keep the interface up to date as the runtime APIs are updated. Thanks!

PR #199, opened 2023-01-04
Fix: Fixes Runtime problem with ESM module

bhelx:
Takes new version of the wasi shim and ensures that both CJS and ESM distributions can be built.

Already published as 0.2.2


browser/package.json
```diff
@@ -17,7 +17,7 @@
   ],
   "module": "dist/index.esm.js",
   "main": "dist/index.js",
-  "typings": "dist/index.d.ts",
+  "typings": "dist/src/index.d.ts",
```
bhelx, line 20: not sure when this changed

PR #201, opened 2023-01-06
feat: add `EXTISM_DEBUG` macro to enable debug info

zshipko:


zshipko:
I tried to look into the browser runtime error but couldn't figure it out, seems unrelated to this change though. /cc @bhelx

bhelx:
@zshipko yes, that's my fault. Didn't think it would stop the builds. feel free to ignore it for now it's a problem with the test not the code.

Issue #202, opened 2023-01-06
Rich `Context` API and decoupling `Plugin`s


ok-nick:
[extism::Context](https://docs.rs/extism/latest/extism/struct.Context.html) has two, very minimal functions for managing a pool of plugins. Underneath the hood, in [extism_runtime::Context](https://docs.rs/extism-runtime/latest/extism_runtime/struct.Context.html), there are numerous useful functions that could be exposed so that the user doesn't need to reimplement it themselves.

I found that I often need to manage a pool of `Plugin`s. To currently do this, you must create your own collection, when `extism_runtime::Context` already does this for you. 

On that note, I think it might be a little nicer to decouple a `Plugin` and a `Context`. You can create a plugin on your own, then when you want to use it, you upload it to a `Context`. I think the reason this wasn't done originally is so that you don't have to pass a `Context` to `Plugin::call`, which is why I also think that  there should be a distinction between a `Plugin` and a `PluginSchema`. Essentially, a `PluginSchema` is a way to upload a WASM module to a `Context` with a given `Manifest` and other additional information. After upload, it will return a reference to the `Plugin` to which you can then call functions on. When you need it for another time, you pull the `Plugin` from the `Context`, using the proper API, then operate on it. This turns a `Context` into a custom collection for managing `Plugin`s.

In addition to the above, it would also be useful to expose `PluginIndex`s and possibly `extism_runtime::Context::next_id`. I found that I want to generate ids for plugins to identify them, however, these plugins are not directly managed by `extism`. Maybe the "unique id" system should be decoupled into a separate struct that is used by a `Context` and could be passed by the user (otherwise defaulted).

Edit: I see the point of why `Context`s aren't a proper collection is because a `Plugin` can be freed after being dropped. That's nice and all, but often times I need to store it and call it later. Why force the user to create their own collection when it's done internally? `extism` can still expose a wrapper struct around a `Plugin` that frees on drop, but I just don't think it's that useful when users can free them by calling a separate function.

zshipko:
Thanks for the suggestions! I think some of these ideas could be nice - a way to move a plugin across contexts could be interesting but it's not really on our roadmap right now - I would be happy to review a PR!

Without that capability a plugin's lifetime depends on the lifetime of the context used to create it - you can manually ensure that the context lives long enough and use `std::mem::forget(plugin)` then `drop(plugin)` but I'm not convinced that should be the default behavior.

ok-nick:
[...]

PR #203, opened 2023-01-06
feat(rust-sdk): `Context` collection

ok-nick:
## Example implementation of the idea proposed in #202.

The goal here is to treat `Context` as it is treated internally; a collection of plugins. This will allow users to store and call plugins whenever they want, without them being freed and without having to cache them again.  

A lot of the code doesn't make use of the libraries conventions, I threw it together as an example. Let me know what you think and if we should go ahead with implementing this.

### DIFF
--- a/rust/src/context.rs
+++ b/rust/src/context.rs
@@ -1,7 +1,7 @@
 use extism_runtime::PluginIndex;
 
 use crate::{
-    plugin_builder::{PluginScheme, Source},
+    plugin_builder::{ Source},
     *,
 };
 
@@ -27,12 +27,17 @@ impl Context {
         todo!()
     }
 
-    pub fn insert(&self, scheme: PluginScheme) -> Result<Plugin, Error> {
-        match scheme.source {
+    pub fn insert(
+        &self,
+        source: Source,
+        functions: Vec<Function>,
+        wasi: bool,
+    ) -> Result<Plugin, Error> {
+        match source {
             Source::Manifest(m) => {
-                Plugin::new_with_manifest_and_functions(&self, &m, scheme.functions, scheme.wasi)
+                Plugin::new_with_manifest_and_functions(self, &m, functions, wasi)
             }
-            Source::Data(d) => Plugin::new_with_functions(&self, &d, scheme.functions, scheme.wasi),
+            Source::Data(d) => Plugin::new_with_functions(self, d, functions, wasi),
         }
     }
 
--- a/rust/src/context.rs
+++ b/rust/src/context.rs
@@ -1,4 +1,9 @@
-use crate::*;
+use extism_runtime::PluginIndex;
+
+use crate::{
+    plugin_builder::{PluginScheme, Source},
+    *,
+};
 
 pub struct Context(pub(crate) std::sync::Arc<std::sync::Mutex<extism_runtime::Context>>);
 
@@ -16,6 +21,25 @@ impl Context {
         )))
     }
 
+    pub fn get(&self, id: PluginIndex) -> Option<&mut extism_runtime::Plugin> {
+        // TODO: an example impl
+        // self.lock().plugin(id)
+        todo!()
+    }
+
+    pub fn insert(&self, scheme: PluginScheme) -> Result<Plugin, Error> {
+        match scheme.source {
+            Source::Manifest(m) => {
+                Plugin::new_with_manifest_and_functions(&self, &m, scheme.functions, scheme.wasi)
+            }
+            Source::Data(d) => Plugin::new_with_functions(&self, &d, scheme.functions, scheme.wasi),
+        }
+    }
+
+    pub fn remove(&mut self, plugin: Plugin<'_>) {
+        unsafe { bindings::extism_plugin_free(&mut *self.lock(), plugin.id) }
+    }
+
     /// Remove all registered plugins
     pub fn reset(&mut self) {
         unsafe { bindings::extism_context_reset(&mut *self.lock()) }
@@ -26,5 +50,11 @@ impl Context {
             Ok(x) => x,
             Err(x) => x.into_inner(),
         }
+    }
+}
+
+impl Drop for Context {
+    fn drop(&mut self) {
+        self.reset();
     }
 }
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -7,7 +7,7 @@ mod plugin_builder;
 
 pub use context::Context;
 pub use plugin::Plugin;
-pub use plugin_builder::PluginBuilder;
+pub use plugin_builder::PluginSchemeBuilder;
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
@@ -150,8 +150,9 @@ mod tests {
 
         std::thread::spawn(|| {
             let context = Context::new();
-            let mut plugin = PluginBuilder::new_with_module(WASM)
-                .build(&context)
+            let scheme = PluginSchemeBuilder::new_with_module(WASM)
+                .build();
+            let mut plugin = context.insert(scheme)
                 .unwrap();
             let output = plugin.call("count_vowels", "this is a test aaa").unwrap();
             std::io::stdout().write_all(output).unwrap();
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -7,7 +7,7 @@ mod plugin_builder;
 
 pub use context::Context;
 pub use plugin::Plugin;
-pub use plugin_builder::PluginSchemeBuilder;
+pub use plugin_builder::PluginBuilder;
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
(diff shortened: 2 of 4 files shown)


ok-nick:
I don't think schemes are necessary, we could probably remove them in favor of the old API, while still reforming the `Context` to be a collection.

zshipko:
Thanks! I agree that treating the `Context` as a collection of plugins is helpful is some cases, but I'm not sure how schemes are different from just using the manifest?

Adding some way to persist a plugin like `Plugin::into_raw(self)` would make it so `Context::get` and `Context::remove` could be added without the need for schemes. I think this would accomplish the same thing:

```rust
impl<'a> Plugin<'a> {
  pub fn into_raw(self) -> PluginIndex {
    let id = self.id;
    std::mem::forget(self);
    id
  }
}
``` 

This allows you to retain the ID of the plugin without dropping it, which means you could then opt-in to using `Context::get` and `Context::remove` to handle plugins.

ok-nick:
@zshipko While that works, it feels very unnatural and unclear, not too Rust-like. Imagine you have a plugin that you want to create but not use until later:
```rust
Plugin::new(...).into_raw();
```
Could you tell by this code that the `Plugin` is going to be cached? We could rename it to `.cache()`, but still, the plugin is already cached even before we called `.cache()`.

How about we introduce a new struct named `TempPlugin` or `PluginHandle` that, when dropped, will free it from the `Context`? It would look something like this:
```rust
Plugin::builder()
    .with_was(true)
    .build_temp(&context)
```

zshipko:
I want to keep the default behavior for `Plugin` as it is to avoid too much API churn.

What about instead of `into_raw` calling it `keep`? I think that's used in the Rust PDK for something similar.

Inverting your `build_temp` example could work too: adding a `PersistentPlugin` and `PluginBuilder::build_persistent` function. But the extra struct feels somewhat unnecessary since the `PluginIndex` kind of already serves as the persistent handle.

ok-nick:
The problem is when you retrieve a `Plugin` from the `Context` you have to call `.keep()` every time.

zshipko:
Ah, that's true. You could add a private field to `Plugin` called `persistent` (or similar) that is used to toggle the behavior of `drop`, it would be set to `true` when the plugin is retrieved using `Context::get` but `false` otherwise.

ok-nick:
You *can* do that but it would be odd to have a function that only works once then serves no purpose.

I think the `build_persistent` is the way to go and we can use traits to avoid repeating ourselves.

How often do people need a plugin temporarily loaded vs persisted? Is it really worth supporting an API that won't be used? In my case, I call the plugins at random times during the programs execution which is why I need them persisted.

zshipko:
The most typical use case is the create the plugin and store it somewhere to call later, then when it goes out of scope it gets destroyed just like any other Rust value. The `Context` is used primarily for error reporting and other book-keeping. Even in the current implementation you can decide exactly when the plugin is removed, there should be no question when that value goes out of scope.

ok-nick:
If `Context` is just an abstraction over multiple plugins, then maybe we should make it optional? Errors could be contained in the plugin itself, then if the user wants a structure to handle multiple plugins they can construct a `PluginGroup`/`Context`. That way a user can opt-in to persisted plugins with a backing storage.

As far as my code, it's just a system that calls all plugins sharing a protocol and aggregates the results. My code works how it is, but I don't see why I'm making a storage on top of a storage, when I can just use the inner one. Not to mention having to create my own id system to keep track of plugins.

zshipko:
That sounds good to me - I would love for `Context` to be optional! 

I think it may require some deeper changes to the runtime but I would be curious to see what that might look like.

ok-nick:
Here are some ways we can return errors to the caller without a `Context`:
1. Return an error code and pass the success in an out parameter. The caller gets the error string via a separate function.
An error would look something like this:
```rust
#[repr(C)]
pub struct CreatePluginResult {
  success: bool,
  result: *mut Plugin,
}

impl Display for CreatePluginResult {
  fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), fmt::Error> { 
    write!(f, "Error creating Plugin")
  }
}
```
With a function to get the error string as such:
```rust
pub fn error_string<T: Display>(error: T) -> *mut c_char {
  todo!() // return display function
}
```
And a macro to make defining errors easier:
```rust
c_error!(
  #[error("Error creating Plugin")]
  CreatePluginResult: *mut Plugin,
)
```
2. Return a C struct containing whether it succeeded or failed along with the result. The only problem with this is that we need a new struct per error which is a little inconvenient, but arguably better. It can probably be minimized using a macro of some such
An error would look something like this:
```rust
#[repr(C, i32)]
pub enum Error {
  CreatePluginResult = 1,
}

impl Display for CreatePluginResult {
  fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), fmt::Error> { 
    match *self {
      Error::CreatePluginResult => write!(f, "Error creating Plugin"),
    }
  }
}
```
The function to get the error string would be the same as the previous and a macro could also be made, though I don't think it'd be necessary. The only difference here that you have to remember is that the result is being passed as a pointer in the parameter.

Another thing is that the library should use a crate like `thiserror` instead of `anyhow`. `anyhow` is meant for binaries and there is no way to know what error happened, only that an error did happen.

zshipko:
Interesting, the idea of returning `*mut Plugin` directly is something I've looked at a few times but we're not looking to switch to thiserror. The out argument could always be used to return an error string to the caller, that is pretty typical in C libraries. 

You have to keep in mind that any change like this will bubble up to all the SDKs - for example, using an out parameter from Javascript is do-able, but it's tricky - there are reasons things are the way they are when you consider how many languages are supported. Yes the API could be nicer, but it part of the design process was making it usable from many languages. 

I would suggest you try to make some of the API changes you're looking for and change a few SDKs to match to get a feeling of how usable the API is from the different languages we support.

ok-nick:
If the language supports FFI, it surely supports out parameters, it's such a widely used C convention. Perhaps it is a little tricky to get right, but that will all be hidden from the user anyways.

Also, the `#[error("blah")]` was not `thiserror` syntax, it was just an example of how we can define a custom macro, I just tried to make it look more idiomatic. Though, on the topic of `thiserror`, I do believe it is something that should eventually be integrated into the library. If not `thiserror`, then at least something to differentiate between errors so that callers can handle accordingly.

**If anyone is familiar with FFI in any of the supported SDKs, feel free to give your thoughts.**

ok-nick:
I came up with an example of what the code would look like with the new error system (1st method).
Here is the macro to make defining errors easier:
```rust
macro_rules! c_error {
    (
        $(
            $error_msg:literal: $error_arg:ty,
            $error_name:ident: $result:ty,
        ),+
    ) => {
        $(
            #[repr(C)]
            pub struct $error_name {
                pub success: bool,
                pub result: $result,
                pub error_arg: $error_arg,
            }

            impl Display for $error_name {
                fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), fmt::Error> {
                    write!(f, $error_msg, self.error_arg)
                }
            }
        ),+
    }
}
```

Creating a new error:
```rust
c_error!(
    "Error creating Plugin: {:?}": Option<anyhow::Error>,
    CreatePluginResult: *mut Plugin,
);
```

What the C function would look like:
```rust
#[no_mangle]
pub unsafe extern "C" fn extism_plugin_new(
    wasm: *const u8,
    wasm_size: Size,
    with_wasi: bool,
) -> CreatePluginResult {
    trace!("Call to extism_plugin_new with wasm pointer {:?}", wasm);
    let data = std::slice::from_raw_parts(wasm, wasm_size as usize);
    match Plugin::new(data, with_wasi) {
        Ok(mut x) => {
            // TODO: forget x
            CreatePluginResult {
                success: true,
                result: &mut x as *mut _,
                error_arg: None,
            }},
        Err(e) => {
            error!("Error creating Plugin: {:?}", e);
            CreatePluginResult {
                success: false,
                result: ptr::null_mut(),
                error_arg: None,
            }
        }
    }
 }
```

I'm not a huge fan of the `error_arg` and it should definitely be abstracted away. Perhaps we can make a builder for each error type or something of the sort.

ok-nick:
Something like this could make creating the errors easier:
```rust
// success
CreatePluginResult::success(x) // convert it to pointer internally?
// error
CreatePluginResult::error(err)
```

nilslice:
@ok-nick - I think there are is more to this PR than meets the eye here, and there's probably more context that would be helpful to know in order for us to really understand whether or not it's something we'd want to accept upstream. 

We created the [EIP repo](https://github.com/extism/proposals) for these kinds of ideas, and if you would like to, I'd encourage you to flesh out these ideas a bit more in the form of a proposal. 

These proposals should ultimately be as thorough as [EIP-005](https://github.com/extism/proposals/blob/main/EIP-005-extism-versioning-system.md), which goes in depth about how all Extism components will be versioned, including the various components, the complexity, and the thoughtful rationale for the versioning system we've landed on. 

I'm going to close out this PR with preference for a proposal, and after that has been accepted, I think everyone involved here will have more clarity about the how/what/why mentioned above. 

Thanks!

PR #208, opened 2023-01-09
Support jdk11

Zwiterrion:
Hi,

We want to use Extism on our project [Otoroshi](https://github.com/MAIF/otoroshi) but we need to run it on jdk11

This pull request makes everything run smoothly on jdk11

If you have any suggestion about this pull request, i'm open to it

Thanks for your time

### DIFF
--- a/.github/workflows/ci-java.yml
+++ b/.github/workflows/ci-java.yml
@@ -28,7 +28,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: '17'
+          java-version: '11'
       - name: Test Java
         run: |
           cd java
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -46,7 +46,7 @@
     </issueManagement>
 
     <properties>
-        <java.version>17</java.version>
+        <java.version>11</java.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
         <!-- dependencies -->
--- a/java/src/main/java/org/extism/sdk/manifest/ManifestHttpRequest.java
+++ b/java/src/main/java/org/extism/sdk/manifest/ManifestHttpRequest.java
@@ -5,10 +5,25 @@ import java.util.Map;
 // FIXME remove this and related stuff if not supported in java-sdk
 public class ManifestHttpRequest {
 
-    private String url;
+    private final String url;
+    private final Map<String, String> header;
+    private final String method;
 
-    private Map<String, String> header;
+    public ManifestHttpRequest(String url, Map<String, String> header, String method) {
+        this.url = url;
+        this.header = header;
+        this.method = method;
+    }
 
-    private String method;
+    public String url() {
+        return url;
+    }
+
+    public Map<String, String> header() {
+        return header;
+    }
+
+    public String method() {
+        return method;
+    }
 }
-
--- a/java/src/main/java/org/extism/sdk/manifest/ManifestHttpRequest.java
+++ b/java/src/main/java/org/extism/sdk/manifest/ManifestHttpRequest.java
@@ -3,6 +3,12 @@ package org.extism.sdk.manifest;
 import java.util.Map;
 
 // FIXME remove this and related stuff if not supported in java-sdk
-public record ManifestHttpRequest(String url, Map<String, String> header, String method) {
+public class ManifestHttpRequest {
+
+    private String url;
+
+    private Map<String, String> header;
+
+    private String method;
 }
 
--- a/java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
+++ b/java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
@@ -9,8 +9,8 @@ import com.google.gson.annotations.SerializedName;
  * @param max Max number of pages.
  */
 public class MemoryOptions {
-
-    private @SerializedName("max") Integer max;
+    @SerializedName("max")
+    private final Integer max;
 
     public MemoryOptions(Integer max) {
         this.max = max;
--- a/java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
+++ b/java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
@@ -8,5 +8,11 @@ import com.google.gson.annotations.SerializedName;
  *
  * @param max Max number of pages.
  */
-public record MemoryOptions(@SerializedName("max") Integer max) {
+public class MemoryOptions {
+
+    private @SerializedName("max") Integer max;
+
+    public MemoryOptions(Integer max) {
+        this.max = max;
+    }
 }
--- a/java/src/main/java/org/extism/sdk/wasm/ByteArrayWasmSource.java
+++ b/java/src/main/java/org/extism/sdk/wasm/ByteArrayWasmSource.java
@@ -2,10 +2,35 @@ package org.extism.sdk.wasm;
 
 /**
  * WASM Source represented by raw bytes.
- *
- * @param name
- * @param data the byte array representing the WASM code
- * @param hash
  */
-public record ByteArrayWasmSource(String name, byte[] data, String hash) implements WasmSource {
+public class ByteArrayWasmSource implements WasmSource {
+
+    private String name;
+
+    private byte[] data;
+
+    private String hash;
+
+
+    /**
+     * Constructor
+     * @param name
+     * @param data the byte array representing the WASM code
+     * @param hash
+     */
+    public ByteArrayWasmSource(String name, byte[] data, String hash) {
+        this.name = name;
+        this.data = data;
+        this.hash = hash;
+    }
+
+    @Override
+    public String name() {
+        return name;
+    }
+
+    @Override
+    public String hash() {
+        return hash;
+    }
 }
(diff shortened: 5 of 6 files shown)


.github/workflows/ci-java.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: '17'
+          java-version: '11'
```
bhelx, line 31: Should we perhaps setup a matrix here and test on both versions, or should we just be testing the lowest version and assuming the latest version is ok? I'm a bit unsure about what the best strategy is here for java.

java/pom.xml
```diff
@@ -46,7 +46,7 @@
     </issueManagement>
 
     <properties>
-        <java.version>17</java.version>
+        <java.version>11</java.version>
```
bhelx, line 49: Does this have an affect on what the compiled code looks like? What is the impact of this?

.github/workflows/ci-java.yml
nilslice: I think having a matrix, consisting of the minimum supported version (v11 with this change) and at least a couple more recent versions (including the most recent version, v17 prior to this change) would be a good idea. The evolution of Java between version 11 and 17 is not something I am familiar with though, so this could be an impossible request. Feedback appreciated!

java/pom.xml
Zwiterrion: I didn't notice any differences between the two builds.

[jdk11.log](https://github.com/extism/extism/files/10381062/jdk11.log)
[jdk17.log](https://github.com/extism/extism/files/10381063/jdk17.log)

.github/workflows/ci-java.yml
nilslice: Yea, I think the array of versions makes sense. I think we'd also need to inject that version from the matrix into the `pom.xml` on CI before running it though, right? Maybe other places too?

bhelx:
Perhaps @thomasdarimont has an opinion on this PR.

I support adding more version to the test matrix. But when it comes to publishing jars it gets unclear to me what the consequences are. I think priorities, in this order, would be:

1. Support as many JVM versions as possible for our users
2. Keep the build and release system as simple as possible
3. Allow us to write the source in the newest language but still target older JVMs.

java/src/main/java/org/extism/sdk/manifest/ManifestHttpRequest.java
```diff
@@ -3,6 +3,12 @@
 import java.util.Map;
 
 // FIXME remove this and related stuff if not supported in java-sdk
-public record ManifestHttpRequest(String url, Map<String, String> header, String method) {
+public class ManifestHttpRequest {
+
+    private String url;
```
thomasdarimont, line 8:
```suggestion
    private final String url;
```

java/src/main/java/org/extism/sdk/manifest/ManifestHttpRequest.java
```diff
@@ -3,6 +3,12 @@
...
+
+    private Map<String, String> header;
```
thomasdarimont, line 10:
```suggestion
    private final Map<String, String> header;
```

java/src/main/java/org/extism/sdk/manifest/ManifestHttpRequest.java
```diff
@@ -3,6 +3,12 @@
...
+
+    private String method;
```
thomasdarimont, line 12:
```suggestion
    private final String method;
```

java/src/main/java/org/extism/sdk/manifest/ManifestHttpRequest.java
```diff
@@ -3,6 +3,12 @@
 import java.util.Map;
 
 // FIXME remove this and related stuff if not supported in java-sdk
-public record ManifestHttpRequest(String url, Map<String, String> header, String method) {
+public class ManifestHttpRequest {
```
thomasdarimont, line 6: I'd make the fields final, create accessor methods and create a constructor for passing the required parameters in.

java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
```diff
@@ -8,5 +8,11 @@
  *
  * @param max Max number of pages.
  */
-public record MemoryOptions(@SerializedName("max") Integer max) {
+public class MemoryOptions {
+
+    private @SerializedName("max") Integer max;
```
thomasdarimont, line 13:
```suggestion
    @SerializedName("max")
    private final Integer max;
```

java/src/main/java/org/extism/sdk/manifest/MemoryOptions.java
```diff
@@ -8,5 +8,11 @@
  *
  * @param max Max number of pages.
  */
-public record MemoryOptions(@SerializedName("max") Integer max) {
+public class MemoryOptions {
```
thomasdarimont, line 11: Create a constructor that assigns the max value.

java/src/main/java/org/extism/sdk/wasm/ByteArrayWasmSource.java
```diff
@@ -2,10 +2,35 @@
...
  * WASM Source represented by raw bytes.
- *
- * @param name
- * @param data the byte array representing the WASM code
- * @param hash
  */
-public record ByteArrayWasmSource(String name, byte[] data, String hash) implements WasmSource {
+public class ByteArrayWasmSource implements WasmSource {
+
+    private String name;
```
thomasdarimont, line 8:
```suggestion
    private final String name;
```

java/src/main/java/org/extism/sdk/wasm/ByteArrayWasmSource.java
```diff
@@ -2,10 +2,35 @@
...
+
+    private byte[] data;
```
thomasdarimont, line 10:
```suggestion
    private final byte[] data;
```

java/src/main/java/org/extism/sdk/wasm/ByteArrayWasmSource.java
```diff
@@ -2,10 +2,35 @@
...
+
+    private String hash;
```
thomasdarimont, line 12:
```suggestion
    private final String hash;
```

java/src/main/java/org/extism/sdk/wasm/PathWasmSource.java
```diff
@@ -2,11 +2,35 @@
...
  * WASM Source represented by a file referenced by a path.
- *
- * @param name
- * @param path
- * @param hash
  */
-public record PathWasmSource(String name, String path, String hash) implements WasmSource {
+public class PathWasmSource implements WasmSource {
+
+    private String name;
```
thomasdarimont, line 8:
```suggestion
    private final String name;
```

java/src/main/java/org/extism/sdk/wasm/PathWasmSource.java
```diff
@@ -2,11 +2,35 @@
...
+
+    private String path;
+
+    private String hash;
```
thomasdarimont, line 12:
```suggestion
    private final String hash;
```

thomasdarimont (changes requested):
A few remarks.

chabala:
@nilslice regarding:
> I think we'd also need to inject that version from the matrix into the pom.xml on CI before running it though, right?

and

> I've added the matrix for version testing in this branch: https://github.com/extism/extism/tree/test-java-versions

> Within that new branch, I've updated the CI to manipulate pom.xml with the version being tested, but not sure if there are other files that need similar changes.

I don't think that's a good idea. Java is backward compatible in the sense that newer JREs can recognize older bytecode, but not vice versa. The `pom.xml` should contain the oldest version the project targets, and the produced JAR tested on multiple JRE versions, not changed on the fly and built multiple times. The JAR built with a pom specifying 17 will absolutely fail on earlier versions as @Zwiterrion noted.

nilslice:
@chabala - 😄 uh oh, my "I don't know Java" is showing... 

I just wanted to make sure we continue to test newer versions as well. I assumed we'd need to update the pom.xml to coincide with the version of Java being used, but if that's not the case then that simplifies things!

chabala:
No worries :laughing: I figured I'd chime in as things looked to be going in a weird direction. I had asked about `why Java 17?` in the [r/java announcement post](https://www.reddit.com/r/java/comments/zqwcqz/comment/j10r3ii/), so I figured I'd check in on how supporting earlier versions was going. I'd support your idea of going as far back as version 8 as well, unless you have other requirements that need newer language features; 8 will be in use for a long time yet.

thomasdarimont:
I suggest using at least Java 11. Java 8 is now almost 10 years old and many Java open source projects have removed support for Java 8.

nilslice:
> I suggest using at least Java 11. Java 8 is now almost 10 years old and many Java open source projects have removed support for Java 8.

This sounds good to me. Unless someone feels very strongly about supporting Java 8, and would like to test & make whatever library changes are needed to support it, I think we can use Java 11 as the minimum supported version.

nilslice:
@Zwiterrion - thank you for your work on this! When you have a minute, would you be able to include the changes suggested by @thomasdarimont? Thanks to my new knowledge from @chabala, the `pom.xml` seems like it's already good as-is in your branch, and we should just add the matrix from my branch to include `[11, 17]` versions in the Java CI workflow. 

What do you think?

Zwiterrion:
@nilslice - It's fine for me, i'll try to include all changes suggested by @thomasdarimont on my branch.

nilslice:
Hi @Zwiterrion - just wanted to check in on this. I looks like you made most of the changes - is my observation correct? We'd love to include this in the upcoming 0.2.0 release, but let me know if there is anything left here or if you would like some help!

bhelx (approved):

PR #221, opened 2023-01-18
ci(java-sdk): Test on multiple JREs

bhelx:


.github/workflows/ci-java.yml
```diff
@@ -28,12 +29,9 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: '11'
+          java-version: '${{ matrix.version }}'
```
bhelx, line 32: I think I need to separate the compile and test phases here because this should bring the older JDK as well right? We just want to test on different JREs

PR #225, opened 2023-01-19
docs: Symlink plugin in example files

bhelx:


ruby/example.rb
```diff
@@ -7,7 +7,7 @@
 # If you do you can create a context with `Extism#new`, example: `ctx = Extism.new`
 Extism.with_context do |ctx|
   manifest = {
-    :wasm => [{ :path => "../wasm/code.wasm" }],
+    :wasm => [{ :path => "./code.wasm" }],
```
nilslice, line 10: should we drop the ./ prefix so it's compatible on windows? if it's at this path I think it's ok to omit

ruby/extism.gemspec
```diff
@@ -27,12 +27,13 @@ Gem::Specification.new do |spec|
       (f == __FILE__) || f.match(%r{\A(?:(?:bin|test|spec|features)/|\.(?:git|travis|circleci)|appveyor)})
     end
   end
+  spec.files.reject! { |f| f.end_with?(".wasm") }
```
bhelx, line 30: Make sure they don't end up in the gem

ruby/extism.gemspec
```diff
@@ -27,12 +27,13 @@ Gem::Specification.new do |spec|
...
   spec.bindir = "exe"
   spec.executables = spec.files.grep(%r{\Aexe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
   # Uncomment to register a new dependency of your gem
-  spec.add_dependency "ffi", ">= 1.0.0"
+  spec.add_dependency "ffi", "~> 1.0"
```
bhelx, line 36: drive by commit, it's the same thing but `>=` is too liberal. This ensures we stay on 1.X but get updates.

bhelx:
Closing because this will be irrelevant when we move hosts

PR #236, opened 2023-01-23
chore: Make memory traces clearer

bhelx:
If you look at the memory traces it's not really clear what the values are:

```
extism_runtime::memory TRACE 2023-01-23T10:05:16.918413-06:00 - store_u64: 656c626169726176 at offset 25
extism_runtime::memory TRACE 2023-01-23T10:05:16.918418-06:00 - store_u8: 20 at offset 33
extism_runtime::memory TRACE 2023-01-23T10:05:16.918423-06:00 - store_u8: 76 at offset 34
extism_runtime::memory TRACE 2023-01-23T10:05:16.918428-06:00 - store_u8: 61 at offset 35
extism_runtime::memory TRACE 2023-01-23T10:05:16.918433-06:00 - store_u8: 6c at offset 36
extism_runtime::memory TRACE 2023-01-23T10:05:16.918438-06:00 - store_u8: 75 at offset 37
extism_runtime::memory TRACE 2023-01-23T10:05:16.918443-06:00 - store_u8: 65 at offset 38
```

It's hard to tell that the value is hex and the offset is base10. This changes u8 values to the form `0x00` where there is always 2 digits and it will always use 16 digits for u64.

### DIFF
--- a/runtime/src/memory.rs
+++ b/runtime/src/memory.rs
@@ -61,7 +61,7 @@ impl PluginMemory {
 
     /// Write byte to memory
     pub(crate) fn store_u8(&mut self, offs: usize, data: u8) -> Result<(), MemoryAccessError> {
-        trace!("store_u8: {data:x} at offset {offs}");
+        trace!("store_u8: offset={offs} data={data:#04x}");
         if offs >= self.size() {
             // This should raise MemoryAccessError
             let buf = &mut [0];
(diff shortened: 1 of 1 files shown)


runtime/src/memory.rs
```diff
@@ -61,7 +61,7 @@ impl PluginMemory {
 
     /// Write byte to memory
     pub(crate) fn store_u8(&mut self, offs: usize, data: u8) -> Result<(), MemoryAccessError> {
-        trace!("store_u8: {data:x} at offset {offs}");
+        trace!("store_u8: offset={offs} data={data:#04x}");
```
bhelx, line 64: it's `04` instead of `02` because rust includes the `0x` characters in the max width.

zshipko (approved):

PR #240, opened 2023-01-24
feat: add helper functions for working with input/output of host functions as bytes

zshipko:


python/extism/extism.py
```diff
@@ -464,6 +464,24 @@ def memory_at_offset(self, offs):
         len = _lib.extism_current_plugin_memory_length(self.pointer, offs)
         return Memory(offs, len)
 
+    def return_bytes(self, output, b: bytes):
+        mem = self.alloc(len(b))
+        self.memory(mem)[:] = b
+        output.value = mem.offset
```
bhelx, line 470: should this be `ouput[0] = `?

bhelx: Nevermind. I understand how it works now

bhelx (dismissed):
Let's get this merged. I'll follow up with the types in python. I think we need them in a few more places.

zshipko:
Oops, I didn't see that last comment and ended up adding the Python types. I will merge once CI passes.

PR #246, opened 2023-01-31
feat: add ability to set host function namespace

zshipko:
- Adds `extism_function_set_namespace` to SDK
- Updates SDKs with host function support to allow setting the namespace for a function

### DIFF
--- a/extism.go
+++ b/extism.go
@@ -115,6 +115,17 @@ func NewFunction(name string, inputs []ValType, outputs []ValType, f unsafe.Poin
 	return function
 }
 
+func (f *Function) SetNamespace(s string) {
+	cstr := C.CString(s)
+	defer C.free(unsafe.Pointer(cstr))
+	C.extism_function_set_namespace(f.pointer, cstr)
+}
+
+func (f Function) WithNamespace(s string) Function {
+	f.SetNamespace(s)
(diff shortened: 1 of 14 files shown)


nilslice:
Cool! This also reminds me we need to update the Runtime APIs page: https://extism.org/docs/concepts/runtime-apis

And should probably have a similar page for Plugin APIs. 

Just a TODO for later.

extism.go
```diff
@@ -115,6 +115,17 @@ func NewFunction(name string, inputs []ValType, outputs []ValType, f unsafe.Poin
 	return function
 }
 
+func (f *Function) SetNamespace(s string) {
+	cstr := C.CString(s)
+	defer C.free(unsafe.Pointer(cstr))
+	C.extism_function_set_namespace(f.pointer, cstr)
+}
+
+func (f Function) WithNamespace(s string) Function {
```
nilslice, line 124: does the method receiver here need to be a pointer type too? wonder or do you know if type checker will fail if its called on a non-pointer Function struct?

nilslice: I see, this is so it can be chained on `NewFunction(...).WithNamespace(...)`.. makes sense

nilslice (approved):

PR #247, opened 2023-02-01
chore: update wasmtime to 6.0, bump extism versions

nilslice:
Unsure if now is the best time to do the `extism` crate version bumps, but I figured they'd need to happen at some point in the near future anyways. Happy to revert if there is any opposition. Also, I added back the local `path` property to the Elixir NIF crate manifest. I want to see if this breaks again in CI, or if we can leave it.

### DIFF
--- a/elixir/native/extism_nif/Cargo.toml
+++ b/elixir/native/extism_nif/Cargo.toml
@@ -11,5 +11,5 @@ crate-type = ["cdylib"]
 
 [dependencies]
 rustler = "0.27.0"
-extism = { version = "0.2.0" }
+extism = { version = "0.2.1", path = "../../../rust", package = "extism" }
 log = "0.4"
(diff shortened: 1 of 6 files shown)


zshipko:
Nice! The issue with the extism_nif crate is in the release workflow. I'm not sure what the best way to toggle that path argument is, but for now it's helpful to make sure the new changes are working with the elixir sdk.

zshipko:
@steve: just updated this to work with wasmtime 6.0.0, I think we could merge this (if you think the other changes on this branch are good to go)

zshipko:
@nilslice: just updated this to work with wasmtime 6.0.0, i think we can merge this if you think the other changes on this branch are good to go

zshipko:
the zig failure looks like its related to some language changes, i can update that in another PR

nilslice:
> the zig failure looks like its related to some language changes, i can update that in another PR

@zshipko, yea I think you are right! sounds good to me :shipit:

nilslice:
I can't approve it though since I opened it initially. Please go ahead and do so at your convenience :)

zshipko (approved):

PR #248, opened 2023-02-02
feat: Add Host Functions support for Java SDK

Zwiterrion:


Zwiterrion:
Hi @bhelx, @nilslice

I'm stuck on implementing JAVA hosts functions. 

I feel like i'm on the right way but i'm not getting the right output when running my plugin with host function.

The core of my host function is:

````java
(ExtismCurrentPlugin plugin,
  LibExtism.ExtismVal[] params,
  LibExtism.ExtismVal[] results,
  JsonElement userData) -> {
    System.out.println(plugin.inputString(params[0]));

    int offs = plugin.alloc(4);
    Pointer mem = plugin.memory();
    mem.write(offs, "test".getBytes(), 0, 4);
    results[0].value.i64 = offs;

    System.out.println("Hello from Java!");
}
````

Here the output of the execution.

````sh
{"count": 4}
Hello from Java!
test
ExtismVal{t=1, value=ExtismValUnion{i32=0, i64=13, f32=0.0, f64=0.0}}
Exit Host function
Plugin output length: 0, output: 
````

At the end of the execution of my host function, i'm able to retrieve the `test` value. 

````sh
... (line 52: HostFunction.java)
System.out.println(LibExtism.INSTANCE.extism_current_plugin_memory(currentPlugin).getString(arraysOfOutputs[0].value.i64));

...

System.out.println(outputs);
System.out.println("Exit Host function");
````

But when the plugin is ending, the result is empty because the `LibExtism.INSTANCE.extism_plugin_output_length` returns 0 as data length while  `LibExtism.INSTANCE.extism_plugin_output_data` returns the expected content. 

I think that I missed something on the implementation, if someone has a idea it's welcome.

**Comment by bhelx** [comment]
Interesting, the output length of 0 seems wrong. I will pull down your branch and take a look today

**Comment by bhelx** [comment]
Got a chance to repro this. It does seem like, yes, the output isn't being registered just for this one test. Going to do a little more testing.

**Comment by bhelx** [comment]
Did a little experimenting with Zach. We narrowed it down to where the problem is. I wrote this little plugin to look at what the host function is returning from the plugin's perspective:

```rust
#![no_main]

use extism_pdk::{*, bindings::extism_alloc};
use serde::{Deserialize, Serialize};

const VOWELS: &[char] = &['a', 'A', 'e', 'E', 'i', 'I', 'o', 'O', 'u', 'U'];

#[derive(Serialize, Deserialize)]
struct Output {
    pub count: i32,
}

extern "C" {
    fn hello_world(input: i64) -> i64;
}

#[plugin_fn]
pub unsafe fn count_vowels<'a>(input: String) -> FnResult<String> {
    let mut count = 0;
    for ch in input.chars() {
        if VOWELS.contains(&ch) {
            count += 1;
        }
    }

    let output = "hello world";
    let memory = Memory::from_bytes(output.as_bytes());
    let output = unsafe { hello_world(memory.offset as i64) };
    Ok(output.to_string())
}

```


It appears `let output` here is always `0`. But logging the outputs in the host function, we can see they are being set. This suggests to us that some kind of pass by value is happening regarding the output array. The original object still has 0 as a default. 

When you set the output value here: https://github.com/extism/extism/pull/248/files#diff-d93dc91be87e544e8ba1116ce83c15697ae8b5c56991f62e57cfb7bf461cefe8R58

you need to ensure you are mutating the reference, and not creating a new value. We looked into `ExtismVal` and `ExtismValUnion` but this has challenged my java knowledge. 

Does that help at all? I think if you can maybe log out object ids and figure out where some value might be being copied, you can find it. If you are still stuck happy to do some more debugging on my end. It is so close!

Zwiterrion:
Thanks for your experiments. I spent the day to update this outputs parameter and it finally works. I'm not expert to mapping C type with JNA and I forgot to define the type used in my union implementation. This oversight prevented rust from getting the correct output value.

For the definition of the host function, I've seen on other sdks that user data is either a pointer or a buffer. I chose to let the user pass JSON as data to be the most generic, but let me know if that's a good idea or not.

bhelx:
@Zwiterrion excellent! Changes work for me too. I will review today.

java/src/main/java/org/extism/sdk/LibExtism.java
```diff
@@ -15,7 +16,105 @@ public interface LibExtism extends Library {
...
+
+        @Override
+        protected List<String> getFieldOrder() {
+            return Arrays.asList("t", "v");
+        }
+
+        public int t;
+        public ExtismValUnion v;
+    }
+
```
bhelx, line 59: Wow this implementation is wild, glad you figured it out 😄

java/src/test/java/org/extism/sdk/PluginTests.java
```diff
@@ -104,4 +110,44 @@ public void shouldAllowInvokeFunctionFromFileWasmSourceMultipleTimesByReusingCon
...
+            System.out.println(String.format("Input string received from plugin, %s", plugin.inputString(params[0])));
+
+            int offs = plugin.alloc(4);
+            Pointer mem = plugin.memory();
+            mem.write(offs, data.map(JsonElement::getAsString).orElse("test").getBytes(), 0, 4);
+            returns[0].v.i64 = offs;
+        };
+
+        String str = "test";
+        Pointer hostUserData = new Memory(str.length() + 1);
```
bhelx, line 129: I think there is a slight misunderstanding about the user data here. it is not something the plugin typically accesses. You can think of it like applied arguments in a closure, or some properties in a class. In this case it's for the java code, and not the plugin  code.

Here is an example of passing in a db connection: https://github.com/extism/extism-sqlite-host-function

here you could pass in things specific to the user. E.G. a session or something.

I'm taking a look at strategies to handle this in Java today.

java/src/main/java/org/extism/sdk/LibExtism.java
Zwiterrion: I searched and tested but in fact it is no longer useful.

java/src/main/java/org/extism/sdk/LibExtism.java
```diff
@@ -19,45 +19,20 @@ public interface LibExtism extends Library {
...
+                ExtismVal inputs,
                 int nInputs,
-                ExtismVal.ByReference outputs,
+                ExtismVal outputs,
                 int nOutputs,
                 Pointer data
         );
     }
 
+    @Structure.FieldOrder({"t", "v"})
```
bhelx, line 30: Oh nice, this is much simpler!

java/src/test/java/org/extism/sdk/PluginTests.java
bhelx: I'm a bit unsure how to make all this work just trying to express the idea right now.

bhelx: With other languages, like python, we have something like a concept of handles: https://github.com/extism/extism/blob/main/python/extism/extism.py#LL205C30-L205C45

this allows us to pass a handle to python objects across the ffi boundary, then when we come back we can retried those python objects from the handle. We need the equivalent thing. I asked ChatGPT what it thinks the equivalent is and it said Callback https://java-native-access.github.io/jna/4.2.1/com/sun/jna/Callback.html

so perhaps we can put some data in a callback and pull it out when we get back into java world?

Zwiterrion:
That's a great idea, I follow your idea (with some tweaks) and it works. I can create a class that inherits from HostUserData with all the fields I want.

````java
class MyUserData extends HostUserData {
      private String data1;
      private int data2;

      public MyUserData(String data1, int data2) {
          super();
          this.data1 = data1;
          this.data2 = data2;
      }
  }

  ExtismFunction helloWorldFunction = (ExtismFunction<MyUserData>) (plugin, params, returns, data) -> {
      System.out.println("Hello from Java Host Function!");
      System.out.println(String.format("Input string received from plugin, %s", plugin.inputString(params[0])));

      int offs = plugin.alloc(4);
      Pointer mem = plugin.memory();
      mem.write(offs, "test".getBytes(), 0, 4);
      returns[0].v.i64 = offs;

      System.out.println(String.format("Host user data, %s, %d", data.data1, data.data2));
  };

  HostFunction hello_world = new HostFunction<>(
          "hello_world",
          parametersTypes,
          resultsTypes,
          helloWorldFunction,
          new MyUserData("test", 2)
  );
````

The HostUserData class just needs to inherit from PointerType, not from structure, and it works!

**Comment by bhelx** [comment]
> The HostUserData class just needs to inherit from PointerType, not from structure, and it works!

That makes a lot more sense! Awesome that it works!

File: java/src/test/java/org/extism/sdk/PluginTests.java [hunk 4]
```diff
@@ -104,4 +110,53 @@ public void shouldAllowInvokeFunctionFromFileWasmSourceMultipleTimesByReusingCon
...
+
+        HostFunction hello_world = new HostFunction<>(
+                "hello_world",
+                parametersTypes,
+                resultsTypes,
+                helloWorldFunction,
+                new MyUserData("test", 2)
+        );
+
+        HostFunction[] functions = {hello_world};
```
bhelx, line 149: I'm not sure the best way to do this in java as my java knowledge is quite outdated, but is it possible to make the generic type optional? So you could pass none or null to the userdata param? Something like this:

```

        HostFunction null_user_data = new HostFunction(
                "hello_world",
                parametersTypes,
                resultsTypes,
                helloWorldFunction,
                null
        );
```

There are definitely cases where a host function doesn't need a user data so it might be cumbersome to have someone make a type. Or maybe we provide some kind of none type for them if there is no generic way to handle it.

bhelx:
@Zwiterrion this is looking good and I'm not seeing any major issues right now. Is it ready to be taken out of draft? Let me know what is left to be done I can maybe pick up some tasks. Or if you think it's good enough at this point we can get it tested and merged.

Zwiterrion:
@bhelx I just replaced the user data by an Java optional. The user can pass an `Optional.empty()` or an `Optional.of(<user defined type>)`. I added two more tests. The first to check the presence of the expected host function hello_world. The second to define a host function without user data.

bhelx:
@Zwiterrion going to review tomorrow. If all is well will get it merged this week! Let me know if there are any last minute changes you want to make.

bhelx:
Alright, just tested again. I think it's in a good spot to merge. If you have any more changes feel free to put in a follow up. We're going to try to release a `0.3` in the coming weeks so there is still some time to make any tweaks or bug fixes.

bhelx (approved):

bhelx:
Thanks so much!

PR #253, opened 2023-02-07
feat: Support for listing exports of a plugin

### LINKED ISSUE #207: FR: Iterable Object for Functions
Hi,

It would be amazing to have an iterable object for the functions available to call on the wasm module.
The online playground does this to populate the dropdown but I believe it requires secondary files. If possible, loading from the  `.wasm` file would be amazing for schema checking and matching functions at runtime. 
Checking if a specific function exists on the wasm object is currently implemented, however creating an iterable opens this up further.

Keep up the great work,
Bowbee

usdogu:
Closes #207 

New functions:
- `extism_plugin_export_list`
- `extism_plugin_export_count`
- `extism_plugin_export_list_free`

I'm by no means a Rust expert, let alone unsafe Rust. So, please review carefully

### DIFF
--- a/cpp/extism.hpp
+++ b/cpp/extism.hpp
@@ -478,6 +478,18 @@ public:
     return extism_plugin_function_exists(this->context.get(), this->plugin,
                                          func.c_str());
   }
+
+  std::vector<std::string> list_exports() const {
+      char **c_exports = extism_plugin_export_list(this->context.get(), this->plugin);
+      int exports_len = extism_plugin_export_count(this->context.get(), this->plugin);
+      std::vector<std::string> exports;
+      for (int i = 0; i < exports_len; i++)
+      {
+        exports.push_back(c_exports[i]);
+      }
+      extism_plugin_export_list_free(c_exports, exports_len);
+      return exports;
+  }
 };
 
 class Context {
--- a/extism.go
+++ b/extism.go
@@ -382,6 +382,21 @@ func (plugin Plugin) FunctionExists(functionName string) bool {
 	return bool(b)
 }
 
+func (plugin Plugin) ListExports() []string {
+	c_exports := C.extism_plugin_export_list(plugin.ctx.pointer, C.int(plugin.id))
+	exports_len := C.extism_plugin_export_count(plugin.ctx.pointer, C.int(plugin.id))
+	exports := make([]string, int(exports_len))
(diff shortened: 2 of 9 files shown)


zshipko:
Thanks, this looks great and will be extremely useful! I will take some time to review it today.

nilslice:
@usdogu - I just met with @zshipko and @bhelx about this and while we think its a good idea to make this easier to do, we think there might be an alternate path that doesn't add more to the Runtime APIs. 

I wanted to say that we really appreciate your time on this work though, and are not opposed to merging it if we all agree its the best path forward. We want to ask your input (and that of @bowbee, who I believe originally made the feature request in #207) about the alternative solution we have in mind. 

We already do deep wasm introspection and analysis in another library, which is _already_ implemented as an Extism plugin here: https://github.com/dylibso/modsurfer/blob/61fe4aee3dfe6601bafb96a7ba6efa5c5106c95e/cli/src/cmd/validate.rs#L280

This function wraps the Extism plugin call of a module with a `parse_module` function. It returns a protobuf encoded [`Module`](https://github.com/dylibso/modsurfer/blob/61fe4aee3dfe6601bafb96a7ba6efa5c5106c95e/module/src/module.rs#L21), which has far more information about the wasm than only the list of exports (which it includes as well). So, instead, by using the [`modsurfer.wasm`](https://github.com/dylibso/modsurfer/blob/61fe4aee3dfe6601bafb96a7ba6efa5c5106c95e/cli/src/plugins/modsurfer.wasm) file as an Extism plugin, and then calling its `parse_module` export with your plugin's wasm bytes, you'd get the exports listed, in addition to even more interesting information. 

We understand the simplicity it adds to have this a part of Extism, but we fear that it starts to bloat the Runtime APIs with surface area that isn't necessarily a core need for Extism itself. I'd ask for you and @bowbee to both share some thoughts if you agree/disagree, and if the Modsurfer library/plugin approach is valid based on the use cases you've thought of.

usdogu:
I was thinking about bloat too while working on this, I definitely agree that we shouldn't pollute the runtime APIs with non-core functions and a plugin like modsurfer makes much more sense

Bowbee:
Seconding @usdogu 's comment here. I agree with not polluting the runtime and keeping it as trim as possible. I think as development continues it may prove useful to have a documentation page of developer recommended useful plugins for users who want more extensibility.

PR #249, opened 2023-02-04
feat(zig-sdk): Implement Host Functions

usdogu:
There are some breaking changes:

- Plugin.init now requires an allocator and functions slice
- Plugin.initFromManifest now requires a functions slice
- Plugin struct removed in favor of using the file as struct
  - from @nilslice: https://zig.news/gowind/zig-files-are-structs-288j (good reference!)

PR #254, opened 2023-02-10
cleanup: use serde(deny_unknown_fields) in manifest

zshipko:


nilslice:
Nice! LGTM, except for maybe flaky Haskell CI. The Zig CI will fail until we merge #249 I think, as it also includes some updated build system changes.

PR #260, opened 2023-02-17
fix(zig-sdk): Fix Example And Test

usdogu:
See ziglang/zig#14638


nilslice (approved):
Awesome - thanks @usdogu! How do you keep on top of these changes to Zig? I'd like to follow more closely as much as I can.

usdogu:
@nilslice, I just check the merged PRs everyday like a psychopath :)

nilslice:
Haha alright I will try to do the same! 

So Zig auto-releases the latest build (and therefore is the same we use in our CI) after these changes are merged?

usdogu:
It auto-releases a new version every day, even if there are no changes

PR #270, opened 2023-03-03
feat: Add plugin cancellation

zshipko:
This PR adds the ability to cancel running plugins from another thread in the host. 

- All SDKs have been updated except Java and .NET
- Adds a new SDK type: `ExtismCancelHandle`
- Adds 2 new SDK functions: `extism_plugin_cancel_handle` and `extism_plugin_cancel`
- `extism_plugin_cancel_handle` returns a pointer to `ExtismCancelHandle`, which can be passed to `extism_plugin_cancel` to stop plugin execution.
- One thing that's worth noting is when plugin is executing a host function it cannot be cancelled until after the host function returns.


Zwiterrion:
Hi @zshipko, I wrote the plugin cancellation for the Java SDK by forking your branch. This is the [PR](#274).

It will be great if you could accept this PR to your branch before doing the next release.

PR #305, opened 2023-04-03
fix(browser): Fix browser runtime, publish as 0.2.7

bhelx:
This upgrades the wasi shim and publishes new version of our runtime. Current published version is broken because it relies on a my git branch which has since been merged into the wasi shim.


bhelx:
Got confirmation that this works

PR #306, opened 2023-04-06
feat: add support for host functions to browser runtime

zshipko:
- Adds ability to define host functions when creating ExtismPlugins in the browser runtime
- The API is a little simpler than the Rust runtime
   - Functions don't handle userdata, userdata should be captured by the function declaration

### DIFF
--- a/browser/index.html
+++ b/browser/index.html
@@ -103,13 +103,13 @@
         }
 
         async loadFunctions(url) {
-            let helloWorld = this.extismContext.newFunction("hello_world", (plugin, index) => {
+            let helloWorld = (index) => {
               console.log("INDEX:" + index);
               return index;
+            };
+            let plugin = await this.extismContext.newPlugin({ "wasm": [ { "path": url } ] }, {
+                "hello_world": helloWorld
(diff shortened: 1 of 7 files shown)


bhelx:
> Functions don't handle userdata, userdata should be captured by the function declaration

so you mean userdata should be captured through scope in the closure?

zshipko:
> > Functions don't handle userdata, userdata should be captured by the function declaration
> 
> so you mean userdata should be captured through scope in the closure?

yeah, I was debating how close it needs to be to the rust runtime. this was much simpler but I could add a userdata argument if you think it would be more usable.

bhelx:
No this makes sense. It’s more idiomatic to JS

zshipko:
@bhelx, @nilslice: were either of you able to test this? I forgot I hadn't merged it yet and don't want to let this get stale!

bhelx:
@zshipko i will test tonight

bhelx:
Looks good to me @zshipko 👍

bhelx (approved):

PR #315, opened 2023-04-21
fix!: disallow http requests by default

zshipko:
- Makes it so requests to a host that isn't specified by `allowed_hosts` will fail
- To allow requests to any URL you can set `{allowed_hosts: ["*"]}`


ubiquitous-dev (approved):
This is an awesome security improvement and definitely in line with the wasm capability philosophy of deny by default.

bhelx:
Added `!` for breaking change

bhelx (approved):

PR #323, opened 2023-04-25
chore: update to wasmtime 8.0.0

zshipko:


bhelx (approved):
This is all we need? looks good to me

PR #326, opened 2023-04-27
feat: add EXTISM_PROFILE environment variable to configure profiling

zshipko:
This could also be extended to support other wasmtime-based profiling methods, for now just `perf` is supported.


nilslice:
Let's add this to the docs somewhere too! Would be a cool thing to demo somehow on the docs site.

bhelx:
We're gonna merge this as it shouldn't cause any harm unless enabled. We can come back to it with additional changes for the next release if we need to.

PR #340, opened 2023-05-10
release: Bump runtime to 0.4.0

bhelx:
## Breaking Changes

* https://github.com/extism/extism/pull/315

HTTP calls will be disallowed by default now. If you want to enable HTTP you need to specify the hosts that the plug-in is allowed to communicate with. If you want to allow all hosts you can set it to `{allowed_hosts: ["*"]}` in the manifest. However, this isn't recommended unless you have some trust in the plug-in or are controlling the networking by some other means.

* https://github.com/extism/extism/pull/335

In this PR we are creating an implicit context so people don't need to know about it if they don't care. In some languages function signatures have changed to make context an optional argument when creating a plug-in.

### DIFF
--- a/dotnet/nuget/Extism.runtime.win.csproj
+++ b/dotnet/nuget/Extism.runtime.win.csproj
@@ -8,7 +8,7 @@
 
     <PropertyGroup>
         <PackageId>Extism.runtime.win-x64</PackageId>
-        <Version>0.5.0</Version>
+        <Version>0.6.0</Version>
         <Authors>Extism Contributors</Authors>
         <Description>Internal implementation package for Extism to work on Windows x64</Description>
         <Tags>extism, wasm, plugin</Tags>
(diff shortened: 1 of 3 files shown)


nilslice:
With #315, I believe we will need to update docs around the default capabilities of the runtime & PDK code. It also obviates the need for the `http` feature, so maybe we should remove that? 

@zshipko can probably shed more light here and help determine what needs to be kept/removed as far as features go.

zshipko:
the docs definitely need to be updated for the runtime and all the affected SDKs. we might want to keep the `http` feature though - compiling without it should result in a smaller runtime since ureq needs rustls and a bunch of other dependencies.

bhelx:
> With https://github.com/extism/extism/pull/315, I believe we will need to update docs around the default capabilities of the runtime & PDK code. It also obviates the need for the http feature, so maybe we should remove that?

In general I wonder about compiler level features for the runtime. Most people experience it as a single shared binary. Asking them to compile it themselves without features seems like a lot. Although I could see someone really concerned about security dipping into that level when going to prod, just to make sure that a runtime / application level change couldn't open up a security hole.

nilslice (approved):
thank you!

PR #341, opened 2023-05-11
fix(rust): use CString for strings passed to to set_log_file

zshipko:


bhelx:
This should work right?

```rust
// host
set_log_file("stdout", Some(log::Level::Trace));
```

```rust
// plugin
info!("Hello World!");
```

Still doesn't seem to be working for me

zshipko:
Does `set_log_file` return `true` or `false`?

bhelx:
It's returning `false`

zshipko:
thanks, i was able to replicate it and just pushed up a fix.

bhelx:
Works now!

bhelx (approved):

PR #344, opened 2023-05-15
fix(php): $err is not implicitly cast to string

bhelx:


nilslice:
@bhelx LGTM, just added the same change in other places where a similar `err` concatenation is done... I think its fine?

nilslice (approved):
LGTM save for the additions I made which you should review... the ol' review ally-oop! :basketball:

PR #348, opened 2023-05-18
fix(ruby): fix implicit context

bhelx:


zshipko (approved):

php/src/Plugin.php
```diff
@@ -31,8 +31,12 @@ class Plugin
 
     private $id;
 
-    public function __construct($ctx, $data, $wasi = false, $config = null) 
+    public function __construct($data, $wasi = false, $config = null, $ctx = null)
```
bhelx, line 34: will get squashed in

PR #356, opened 2023-05-25
fix: require modules to have exported, bounded memory when manifest `memory.max_pages` field is set

zshipko:
- Requires modules compiled to run with manifests that set `max_memory` to have an exported memory with lower and upper bounds
- Includes the size of memory exported from modules when calculating available memory for plugins

## How to compile a module with bounded memory 

You will need to pass `--max-memory=$NUM_BYTES` to wasm-ld. `$NUM_BYTES` must be a multiple of the page size. Here are some examples for supported PDK languages:

**C** 
Pass `-Wl,--max-memory=65536` to your C compiler

**Rust**: 
In a `.cargo/config` file:
```toml
[target.wasm32-unknown-unknown]
rustflags = ["-Clink-args=--max-memory=65536"]
 ```
**Haskell**
Add the following to the cabal file entry for your `cabal.project` file:

```
package myproject
  ghc-options:
    -optl -Wl,--max-memory=65536
```
**AssemblyScript**
Pass `--maximumMemory 65536` to the assemblyscropt compiler

**TinyGo**:
Create a `target.json` file:
```json
{
    "inherits": [ "wasm" ],
    "ldflags": [
        "--max-memory=65536",
    ]
}
```
and build using `tinygo -target ./target.json`

### DIFF
--- a/runtime/src/plugin.rs
+++ b/runtime/src/plugin.rs
@@ -55,6 +55,37 @@ fn profiling_strategy() -> ProfilingStrategy {
     }
 }
 
+fn calculate_available_memory(
+    available_pages: &mut Option<u32>,
+    modules: &BTreeMap<String, Module>,
+) -> anyhow::Result<()> {
+    let available_pages = match available_pages {
+        Some(p) => p,
+        None => return Ok(()),
+    };
+
+    for (name, module) in modules.iter() {
+        let mut memories = 0;
+        for export in module.exports() {
+            if let Some(memory) = export.ty().memory() {
+                let memory_max = memory.maximum();
+                match memory_max {
+                    None => anyhow::bail!("Unbounded memory in module {name}, when `memory.max_pages` is set all modules\
+                                           must have a maximum bound set on an exported memory"),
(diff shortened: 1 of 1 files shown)


runtime/src/plugin.rs
```diff
@@ -55,6 +55,43 @@ fn profiling_strategy() -> ProfilingStrategy {
...
+        for export in module.exports() {
+            if let Some(memory) = export.ty().memory() {
+                let memory_max = memory.maximum();
+                match memory_max {
+                    None => anyhow::bail!("Unbounded memory in module {name}, when `memory.max_pages` is set all modules \
+                                           must have a maximum bound set on an exported memory"),
+                    Some(m) => {
+                        *available_pages = available_pages.saturating_sub(m as u32);
+                        if *available_pages == 0 {
+                            anyhow::bail!("Not enough memory configured to run the provided plugin");
```
bhelx, line 78: Could we perhaps provide some more hints in these error messages? Maybe let the user know they should look at the manifest max_pages config?

bhelx (approved):

PR #375, opened 2023-06-20
chore: use a range for wasmtime dependency version

zshipko:
Currently wasmtime `8.0.0` - `10.0.0` are compatible with `extism-runtime`, in the interest of remaining compatible with as many applications as possible, this PR updates the wasmtime version requirement from a single version to a range of acceptable versions.

### DIFF
--- a/ocaml/manifest/extism_manifest.ml
+++ b/ocaml/manifest/extism_manifest.ml
@@ -1,3 +1,5 @@
+open Ppx_yojson_conv_lib.Yojson_conv
+
 type base64 = string
 
 let yojson_of_base64 x = `String (Base64.encode_exn x)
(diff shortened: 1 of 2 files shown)


ocaml/manifest/extism_manifest.ml
```diff
@@ -1,3 +1,5 @@
+open Ppx_yojson_conv_lib.Yojson_conv
```
bhelx, line 1: Is this meant to be here?

bhelx (approved):
Good to go other than that one question.

ocaml/manifest/extism_manifest.ml
zshipko: Yeah, CI failed without that

PR #384, opened 2023-07-07
feat: Implement parts of the extism runtime in WebAssembly

zshipko:
This PR adds the `kernel` directory which contains a port of the Extism memory allocator compiled to WebAssembly and removes `runtime/src/memory.rs` completely.

Being able to re-use memory functions as a WASM module allows us to begin to experiment with porting Extism to new runtimes!

This is in a draft state while I'm verifying some of these changes.


bhelx:
Could you add a README to the kernel project with a loose description of why it exists? Any graphs or art you have might help too. Either that or a top level module comment.

zshipko:
> Could you add a README to the kernel project with a loose description of why it exists? Any graphs or art you have might help too. Either that or a top level module comment.

Good call! Just added a README: https://github.com/extism/extism/blob/fd783357547c5065284efaaa2755bb517ace7bfa/kernel/README.md and some more documentation to the implementation too: https://github.com/extism/extism/blob/fd783357547c5065284efaaa2755bb517ace7bfa/kernel/src/lib.rs

runtime/src/plugin.rs
```diff
@@ -252,35 +248,55 @@ impl Plugin {
...
+            let linker = &mut self.linker as *mut _;
+            let internal = self.internal_mut();
+            internal.store = store;
+            internal.linker = linker;
+        }
+
+        let bytes = unsafe { std::slice::from_raw_parts(input, len) };
+        trace!("Input size: {}", bytes.len());
+
+        self.start_timer(&tx)?;
```
nilslice, line 276: is this related to the cancellation/timeout we intend to implement in the SDKs now?

runtime/src/plugin.rs
```diff
@@ -290,24 +306,34 @@ impl Plugin {
...
+            self.store = Store::new(
+                &engine,
+                Internal::new(
+                    internal.manifest.clone(),
+                    internal.wasi.is_some(),
+                    internal.available_pages,
+                )?,
+            );
+            self.store
+                .epoch_deadline_callback(|_internal| Err(Error::msg("timeout")));
```
nilslice, line 321: related to the above - just curious if this is something we were going to remove & have implemented elsewhere.

runtime/src/plugin.rs
```diff
@@ -378,22 +401,22 @@ impl Plugin {
     }
 
     pub(crate) fn initialize_runtime(&mut self) -> Result<(), Error> {
```
nilslice, line 403: does this need to be done in kernel-based SDKs?

runtime/src/plugin_ref.rs
```diff
@@ -45,12 +41,25 @@ impl<'a> PluginRef<'a> {
...
+            let plugin = unsafe { &mut *plugin };
+            // Start timer
+            if let Err(e) = plugin.start_timer(&epoch_timer_tx) {
+                let id = plugin.timer_id;
+                plugin.error(
+                    format!("Unable to start timeout manager for {id}: {e:?}"),
+                    (),
+                );
+                return None;
+            }
```
nilslice, line 54: also related to the timeout / cancellation comments above

runtime/src/plugin_ref.rs
```diff
@@ -78,6 +88,14 @@ impl<'a> AsMut<Plugin> for PluginRef<'a> {
...
         trace!("Dropping PluginRef {}", self.id);
-        // Cleanup?
+        if self.running {
+            let plugin = self.as_mut();
+
+            // Stop timer
+            if let Err(e) = plugin.stop_timer() {
+                let id = plugin.timer_id;
+                error!("Failed to stop timeout manager for {id}: {e:?}");
+            }
```
nilslice, line 98: also related to the timeout comments above (sorry if this is redundant or not actually related :laughing:)

runtime/src/sdk.rs
```diff
@@ -574,18 +554,10 @@ pub unsafe extern "C" fn extism_plugin_call(
...
-        return plugin.error(
-            format!("Failed to stop timeout manager for {id}: {e:?}"),
-            -1,
-        );
-    }
-
     match res {
         Ok(()) => (),
         Err(e) => {
+            plugin.store.set_epoch_deadline(1);
```
nilslice, line 560: figuring all of this _is_ actually expected to remain here... I suppose this is at minimum related to the Rust SDK then, and possibly does need to be implemented in kernel-based SDKs?

nilslice (commented):
LGTM, just curious about the timeout/deadline/cancellation things referred to in other comments. 

Really excited about this! 

I think you should self-approve this one when ready ;)

runtime/src/sdk.rs
zshipko: Yeah, timeouts and memory limits are features that will have to be implemented by each kernel-based SDK. This PR doesn't really change anything, but it does make it possible to start building up new SDKs that don't rely on this code anymore. Like in Go, cancellation suddenly becomes way easier because of Contexts, but in Rust it's kinds of a pain right now. If we decided to have an async runtime it would probably change the design a little bit - those are all things we should definitely explore!

runtime/src/plugin.rs
zshipko: For now yeah, that just makes it possible to side-step `_start` in modules that require initialization.

nilslice: Ok, good to know -- I think we need to include it in TODOs in the `go-sdk` cc/ @mhmd-azeez, unless its already being handled!

zshipko: I think we can use `ModuleConfig.WithStartFunctions` to do the same thing in wazero

mhmd-azeez: I added a TODO in the PR readme for it so that we don't forget it

zshipko:
Looks like I can't approve my own PR, but I'm merging this now!

PR #407, opened 2023-08-04
fix: avoid timeout errors outside of `extism_plugin_call`

zshipko:
- Delays instantiation to right before a call by using `instantiate_pre` instead. This fixes an issue with the assemblyscript PDK.
- Makes timeouts only apply to calls
- Also bumps the wasmtime lower-bounds to 10.0, because the return type of the epoch callback has changed


bhelx (approved):
Looks good. by the time we ship this i think people should have time to upgrade to wasmtime 10.

PR #411, opened 2023-08-07
refactor: Simplify runtime handling

zshipko:
It seems like our runtime initialization process is a little too aggressive, this PR scales it back to initialize the runtime once if it exists and only reinitializes when `_start` is called. This prevents globals from being wiped out between plugin calls.

- Removes Runtime::cleanup
- Only initializes runtime once, or if `_start` is called
- Improves Haskell reactor support

See https://github.com/extism/go-sdk/pull/11 for the go-sdk PR.


runtime/src/plugin.rs
```diff
@@ -359,15 +361,21 @@ impl Plugin {
         // Initialize Haskell runtime if `hs_init` and `hs_exit` are present,
```
nilslice, line 361: nit: but we mention hs_exit here in the comment still though, remove the use of that function.

nilslice (dismissed):
Just a note, no problem leaving it in as-is tho!

runtime/src/plugin.rs
zshipko: Thanks1

PR #413, opened 2023-08-11
fix(rust): ensure InternalExt is exported

zshipko:
Also adds a test that uses `InternalExt` in the host function


bhelx (approved):
Worked on my machine, thanks for the updated test too!

PR #414, opened 2023-08-11
refactor(zig)!: Simplify JSON stringifying

usdogu:
Uses `std.json.ArrayHashMap` for stringifying.

# Breaking Changes
* Plugin.setConfig's `config` parameter's type changed to `std.json.ArrayHashMap([]const u8)`
* WasmUrl's `header` field's type changed to `std.json.ArrayHashMap([]const u8)`
* Manifest's `config` and `allowed_path` fields' types changed to `std.json.ArrayHashMap([]const u8)`


zshipko:
Looks great, thanks!

zshipko (approved):

PR #421, opened 2023-08-18
refactor!: Remove context, unify `extism-runtime` and `extism` crates

zshipko:
- Removes the `ExtismContext` type from runtime and all SDKs
  - Removed SDK functions: `extism_context_new`, `extism_context_reset`, `extism_context_free`
  - All SDKs have been updated, but there are still some TODOs below 
- Removes `extism_plugin_update`
  - Plugins can no longer be updated - a new plugin should be created instead
- Adds `extism_plugin_id` to uniquely identify plugins
- Merges the `extism-runtime` and `extism` crates (there is no longer an `extism-runtime` crate)
- Makes `extism::Manifest` an alias for `extism_manifest::Manifest` instead of a distinct type
- Adds `MemoryHandle` type to SDKs to refer to blocks of Extism memory that can be accessed in host functions
- Improves thread-safety of Plugins, adds C++ test to call a single plugin from multiple threads.
- Expands wasmtime bounds to include 12.0

**TODO:**
- [x] Dotnet - Handle error from `extism_plugin_new`, I think I need some help to understand how to make this work. 
- [x] NodeJS -  Same as above, but seems a little trickier, it might make sense to just return a generic error and work on the js-sdk
- [x] Elixir - I was trying to update the Elixir SDK but was having some issues getting it to build locally. It seems to just hang? I will check CI.
- [x] Add `extism_plugin_id` to more SDKs (will follow up with more in a future PR)


dotnet/src/Extism.Sdk/Plugin.cs
```diff
@@ -6,51 +6,43 @@ namespace Extism.Sdk.Native;
...
-
-        var functions = _functions.Select(f => f.NativeHandle).ToArray();
-        fixed (byte* wasmPtr = wasm)
+        unsafe
         {
-            return LibExtism.extism_plugin_update(_context.NativeHandle, Index, wasmPtr, wasm.Length, functions, 0, withWasi);
+            fixed (byte* wasmPtr = wasm)
+            fixed (IntPtr* functionsPtr = functionHandles)
+            {
+                NativeHandle = LibExtism.extism_plugin_new(wasmPtr, wasm.Length, functionsPtr, functions.Length, withWasi, null);
```
zshipko, line 36: @mhmd-azeez - the `extism_plugin_new` function now takes a final parameter which is `char **errormsg`,the way other SDKs have been doing this is to pass a reference to `char*` - I'm not sure how this should be done in C#, do you have any ideas?

mhmd-azeez: Same thing works in dotnet too:

```
_functions = functions;
var functionHandles = functions.Select(f => f.NativeHandle).ToArray();

unsafe
{
    fixed (byte* wasmPtr = wasm)
    fixed (IntPtr* functionsPtr = functionHandles)
    {
        char** errmsg = null;
        NativeHandle = LibExtism.extism_plugin_new(wasmPtr, wasm.Length, functionsPtr, functions.Length, withWasi, errmsg);
        if (NativeHandle == null)
        {
            string error = "Unable to create plugin";
            if (errmsg != null)
            {
                IntPtr err = (IntPtr)(*errmsg);
                error = Marshal.PtrToStringUTF8(err);
                LibExtism.extism_plugin_new_error_free(*errmsg);
            }

            throw new ExtismException(error);
        }
    }
}

```

Pushed a commit has the changes

zshipko: Oh great! Thank you

chances:
Out of curiosity, what is the justification for removing `extism_plugin_update`?

zshipko:
> Out of curiosity, what is the justification for removing `extism_plugin_update`?

It mostly existed to update a Plugin inside of a Context while keeping a stable ID in that context. I'm not sure it was that useful to begin with but I could be convinced to keep something similar if you need it - but I am also happy to reduce the API surface a little bit.

PR #446, opened 2023-09-05
feat: Add `EXTISM_ENABLE_WASI_OUTPUT` to inherit stdout/stderr

zshipko:
@mhmd-azeez pointed out that exposing an option like this can be very useful for debugging if the program is expecting to run in a full WASI environment.


mhmd-azeez (approved):
Perfect!

PR #447, opened 2023-09-05
fix: toml manifests load the extism runtime

chrisdickinson:
Just getting my feet wet with the codebase a little bit! I noticed that TOML manifests weren't loading the extism runtime by default while doing a walkthrough. This commit ensures the runtime is loaded and adds a test.

Also, fix a tiny typo in a comment.


zshipko (approved):
Thanks, definitely appreciate the test!

PR #452, opened 2023-09-12
feat(dotnet): add support for Manifest

mhmd-azeez:
- Add support for manifest
- Add more unit tests
- Update Runtime package reference


zshipko:
There isn't currently a way to access vars from the host, we could add the ability but I don't have any work planned for that yet.

mhmd-azeez:
Will open another PR in the new repo

PR #465, opened 2023-09-18
fix(kernel): avoid allocating 0-length blocks

zshipko:
Fixes an issue where the kernel was able to allocate empty blocks, causing some weird memory issues.


chrisdickinson (approved):
LGTM!

PR #466, opened 2023-09-18
fix: Fixes rounding issue in kernel num_pages

bhelx:
There were some scenarios where the kernel was not allocating enough pages to get data back into the plugin. So some host functions were failing when the output was right on this boundary.

Related https://github.com/moonrepo/proto/issues/208

We will come back with tests and cherry pick this over to main. For now we want to get out a 0.5.1 fix


chrisdickinson (approved):
LGTM!

PR #471, opened 2023-09-19
fix(stable): improve the way the kernel calculates how many pages to allocate

zshipko:
Fixes a bug in the kernel memory allocator where the space used by the `MemoryRoot` wasn't being considered


wikiwong (approved):

rust/src/lib.rs
```diff
@@ -319,4 +320,30 @@ mod tests {
...
+            None,
+            hello_world,
+        );
+
+        let context = Context::new();
+        let mut plugin = Plugin::new(&context, REFLECT_WASM, [f], true).unwrap();
+
+        for i in 1..65540 {
+            let input = "a".repeat(i);
+            println!("i: {i}");
```
bhelx, line 340:
```suggestion
```

rust/src/lib.rs
```diff
@@ -319,4 +320,30 @@ mod tests {
...
+            let output = plugin.call("reflect", input.clone());
+            if let Err(e) = &output {
+                println!("output: #{}", e);
+            }
```
bhelx, line 344:
```suggestion
```

kernel/src/lib.rs
```diff
@@ -121,7 +121,7 @@ pub struct MemoryBlock {
 pub fn num_pages(nbytes: u64) -> usize {
     let npages = nbytes / PAGE_SIZE as u64;
     let remainder = nbytes % PAGE_SIZE as u64;
-    if remainder != 0 || nbytes < PAGE_SIZE as u64 {
```
wikiwong: We no longer need to check if `nbytes < PAGE_SIZE` ?

zshipko: Yeah, @chrisdickinson pointed out in another comment that it's actually alright: https://github.com/extism/extism/pull/472#discussion_r1330777361

PR #472, opened 2023-09-19
fix(main): improve the way the kernel calculates how many pages to allocate

zshipko:
Same as #471 for `main` branch


kernel/src/lib.rs
```diff
@@ -121,7 +121,7 @@ pub struct MemoryBlock {
 pub fn num_pages(nbytes: u64) -> usize {
     let npages = nbytes / PAGE_SIZE as u64;
     let remainder = nbytes % PAGE_SIZE as u64;
-    if remainder != 0 {
+    if remainder != 0 || nbytes < PAGE_SIZE as u64 {
```
chrisdickinson, line 124: We may be able to drop `nbytes < PAGE_SIZE as u64` here: [`16 % 65536` should be `16`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f5a8105670e917110d04a38d089a988b), so the `remainder != 0` check _should_ handle the case where we're requesting fewer bytes than a full page.

kernel/src/lib.rs
```diff
@@ -252,7 +252,7 @@ impl MemoryRoot {
         // we will need to try to grow the memory
         if length >= mem_left {
             // Calculate the number of pages needed to cover the remaining bytes
-            let npages = num_pages(length);
+            let npages = num_pages(length - mem_left);
```
chrisdickinson, line 255: (Ooh, good catch!)

kernel/src/lib.rs
zshipko: Yeah, this doesn't seem to be the issue. I will close it and include the other change in another PR.

PR #481, opened 2023-09-22
fix(build): remove windows gnu wheel builds

chrisdickinson:
Maturin on Windows struggles with GNU headers. Since we have windows MSVC wheel builds, this commit disables windows GNU wheel builds.

Also: fix a bug where the produced wheels did not include any extism functions.

### DIFF
--- a/extism-maturin/pyproject.toml
+++ b/extism-maturin/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "maturin"
 
 [project]
 name = "extism-sys"
-version = "0.0.0-replaced-by-ci"
+version = "0.0.0+replaced-by-ci"
 requires-python = ">=3.7"
 classifiers = [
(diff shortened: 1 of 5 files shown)


extism-maturin/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "maturin"
 
 [project]
 name = "extism-sys"
-version = "0.0.0-replaced-by-ci"
+version = "0.0.0+replaced-by-ci"
```
chrisdickinson, line 7: This makes the placeholder version string compatible with pyproject. Maturin does not accept the hyphenated version; in order to develop locally one would have to replace the version out and remember not to commit the change.

zshipko (approved):
Looks good! I don't see anything that looks particularly strange, but it's hard to tell with these types of changes.

PR #483, opened 2023-09-22
chore: bump wasmtime lower bound to 13.0.0, remove wasi-nn support

zshipko:
- Updates to wasmtime 13.0.0 as the lowest supported version
- Removes `wasmtime-wasi-nn` support because it now takes parameters to specify particular backends and registries during initialization. Since I don't think anyone is using the `nn` feature I chose to remove it for now, but we could also expand the manifest to add these options.

### DIFF
--- a/libextism/Cargo.toml
+++ b/libextism/Cargo.toml
@@ -18,7 +18,6 @@ extism = {path = "../runtime"}
 
 [features]
 default = ["http", "register-http", "register-filesystem"]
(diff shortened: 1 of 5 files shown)


zshipko:
It looks like 13.0.0 breaks a few things so we will have to bump the lower bounds to 13.0.0 - I will update this PR

bhelx (approved):

PR #493, opened 2023-10-02
docs: Add runtime readme

bhelx:


zshipko (commented):
Good idea - made a few notes about how the API has changed since Context has been removed, but this is a good opportunity to update it. I can go through and make these updates - it's mostly just a matter of removing the `ExtismContext*` parameters and using `ExtismPlugin*` instead of `ExtismPlugin`.

bhelx:
Yeah a lot of this is wrong after the changes to 1.0. This is a good example why the source of the doc should be close to the code though. Will update everything and reach out tomorrow.

zshipko:
Just pushed a change to this, it should be up to date now!

zshipko:
Closing in favor of #526 - the API docs have been copied to `libextism/API.md`

PR #494, opened 2023-10-04
chore: prepare `extism-convert` to be released

zshipko:


.github/workflows/release-manifest.yaml
```diff
@@ -0,0 +1,26 @@
+on:
```
bhelx, line 1: Nice!  much cleaner to separate these

bhelx (approved):

PR #497, opened 2023-10-05
chore: fix elixir release action

bhelx:
This is 404ing for some reason. But Elixir doesn't need the shared library so i'm removing it.

https://github.com/extism/extism/actions/runs/6425026823/job/17446865267


zshipko (approved):
Looks like the old CLI, probably why it was failing.

chrisdickinson (approved):

Issue #504, opened 2023-10-10
Update namespace of Extism functions


zshipko:
Currently we use the `env` namespace and prefix our functions with `extism_`, we should be using a proper namespace, like `extism:host/env` instead. So instead of `env::extism_alloc`, we could use `extism:host/env::alloc`. 

For user-defined host functions we should defined to `extism:host/user`

- [x] Update EIP: https://github.com/extism/proposals/pull/14
- [x] Update PDKs 
  - [x] C: https://github.com/extism/c-pdk/pull/7
  - [x] Rust: https://github.com/extism/rust-pdk/pull/36
  - [x] Go: https://github.com/extism/go-pdk/pull/19
  - [x] Dotnet: https://github.com/extism/dotnet-pdk/pull/16
  - [x] Haskell: https://github.com/extism/haskell-pdk/pull/6
  - [x] Zig: https://github.com/extism/zig-pdk/pull/14
  - [x] JS PDK: https://github.com/extism/js-pdk/pull/23
  - [x] Assemblyscript PDK: https://github.com/extism/assemblyscript-pdk/pull/13
- [x] Update SDKs (only listing ones that have references to the `env` namespace)
  - [x] Java: https://github.com/extism/java-sdk/pull/11
  - [x] Haskell: https://github.com/extism/haskell-sdk/pull/11
  - [x] Python: https://github.com/extism/python-sdk/pull/13
- [x] Update runtime: https://github.com/extism/extism/pull/511
- [ ] Update JS-SDK
  - [ ] Update test wasm modules 
- [x] Update Go-SDK: https://github.com/extism/go-sdk/pull/31
  - [x] Update test wasm modules (depends on the new go-pdk changes) 
- [x] Build new test plugins in https://github.com/extism/plugins (might need to wait for Rust PDK to be released?)
  - [x] Updated here: https://github.com/extism/plugins/pull/2 
- [x] Make sure to document the change and that all plugins will need to be rebuilt after this is merged

## Plan
- [x] Merge Go-SDK changes, test the PDK PRs with an update extism CLI
- [x] Merge the PDKs, plugins
  - [x] Dotnet PDK
  - [x] JS PDK 
- [x] Merge the runtime
- [x] Merge the SDKs
- [ ] Update js-sdk
- [x] Merge EIP

PR #511, opened 2023-10-13
refactor!: Use `extism:host/env` namespace for extism functions and `extism:host/user` for user-defined host functions

zshipko:
See #504 

Removes CI for the in-repo SDKs, we can remove the actual code closer to the 1.0 release.


zshipko:
Since the SDKs have mostly been moved to new repos, I'm not going to update the failing SDK tests. Instead, maybe this PR should include the changes to remove the old SDKs from this repo, since we will likely merge this closer to the planned 1.0 release date.

runtime/src/plugin.rs
```diff
@@ -2,6 +2,9 @@ use std::collections::BTreeMap;
 
 use crate::*;
 
+pub const EXTISM_ENV_MODULE: &str = "extism:env";
+pub const EXTISM_USER_MODULE: &str = "extism:user";
```
chrisdickinson, line 6: *winces* Ack, sorry sorry. I think we might want to add an interface projection here, too: something like `extism:core/env` / `extism:core/user`. Mea culpa here – I ran into this with the [observe-api](https://github.com/dylibso/observe-sdk/pull/128/files#diff-0a8d99077aa2dacc5aa5e7d571411c1f496737d0707f978771f9d65cd0674469R3) component model support. (The `/foo` bit maps to an interface in WIT.)

zshipko: Thanks for catching this!!! I think I will go with `extism:host/env` and `extism:host/user`

chrisdickinson (dismissed):
LGTM!

Issue #486, opened 2023-09-25
Bundling libextism


wikiwong:
- Runtimes where this won’t be an issue:
    - Rust (wasmtime)
    - Go (wazero)
    - Javascript (after native Extism runtime is finished)
- Some languages will still depend on libextism being installed as a separate step
    - C/C++
    - OCaml
- Elixir
    - already uses rustler
- Python
    - maturin: https://github.com/extism/python-sdk-maturin
    - Already done
- Haskell
    - cargo-cabal: https://github.com/extism/haskell-sdk
    - Already done
- Ruby
    - https://github.com/matsadler/magnus
    - https://github.com/danielpclark/rutie
    - Estimate: could take 1-3 days to get working, maybe less for someone familiar with Ruby
- Dotnet
    - The maturin CI builds for a bunch of different architectures to makes a release, maybe we can do something similar to bundle with the release.
- Java?
- Zig?

chrisdickinson:
A quick update on `python-sdk`: this is in-progress! Since the native bindings change at roughly the same rate as the main repo, we're building bdist wheels [for release](https://github.com/extism/extism/releases/tag/latest) from `extism/extism` now. These bdist wheels comprise an `extism_sys` package, containing low level `ffi` and `lib` submodules. The higher-level `python-sdk` `extism` package can rely on this binary distribution & provide a higher-level interface using type-hints, etc, released at its own rate.

bhelx:
Will be moving these to individual repos. Example: https://github.com/extism/java-sdk/issues/3

neuronicnobody:
Marking this for closure due to individual repo issues

PR #523, opened 2023-10-17
ci(dotnet): publish nuget packages for all supported environments

mhmd-azeez:
NuGet supports the distribution of native binaries. We already provide a win-x64 package, this PR adds packages for the other environments. We will be publishing these NuGet  packages:
 - Extism.runtime.linux-arm64
 - Extism.runtime.linux-musl-arm64
 - Extism.runtime.linux-x64
 - Extism.runtime.osx-arm64
 - Extism.runtime.osx-x64
 - Extism.runtime.win-x64 (msvc)
 - Extism.runtime.all -> this is a meta package that referencs all other packages. The idea is, in the getting started guides we instruct the user to take a dependency on `Extism.Sdk` and `Extism.runtime.all`. This will make sure their apps work on all supported operating systems automatically. Otherwise, they can just take a dependency on what they care about

This PR also adds a dependency on [MinVer](https://github.com/adamralph/minver) which make sure packages are automatically versioned based on git tags.

 Related to #486


chrisdickinson (approved):
Great work! Landing this and following up by splitting the `dotnet pack` creation from the `dotnet nuget push` steps is perfect.

bhelx (approved):

PR #526, opened 2023-10-18
doc: add `runtime/README.md` and `libextism/README.md` 

zshipko:
C: https://github.com/extism/extism/blob/new-readmes/libextism/README.md
Rust: https://github.com/extism/extism/blob/new-readmes/runtime/README.md

Also includes some small usability improvements I stumbled across when working on the docs


zshipko:
Merging this to get it out of the way to the namespace changes, we can make more doc updates in future PRs if needed.

Issue #537, opened 2023-10-23
feat: Add ability to dump memory to a file when an error occurs


zshipko:
We should add an environment variable like `EXTISM_COREDUMP` to enable memory to be dumped to a file on disk when an error occurs

PR #539, opened 2023-10-24
feat: add ability to dump extism kernel memory and generate coredumps, cleanup kernel memory layout

zshipko:
Fixes https://github.com/extism/extism/issues/537

- Requires wasmtime 14.0.0 or greater for coredump serialization
- Adds `EXTISM_COREDUMP` environment variable to write wasmtime coredump to a file when an error occurs
  - This will create a coredump from the main wasm module, which means it doesn't give us too much insight into the kernel 
- Adds `EXTISM_MEMDUMP` environment variable to write extism linear memory to a file when an error occurs
  - This gives us access to the kernel memory  
- Adds some missing profiling options
- Converts timeouts to a Trap instead of a plain error, this helps us get better information about where the timeout occured
- Some small improvements to error handling after a plugin call
- Adds a test for coredump and memdump generation
- Adds the ability to configure debug options using `PluginBuilder`
- Fixes memory layout and a wasted page of memory in the kernel, found while debugging a memory dump


bhelx (approved):
This is excellent !

bhelx (approved):

PR #548, opened 2023-10-25
fix(release): bump maturin manylinux floor  for ring 0.17 dep

chrisdickinson:
An update to `ring` broke the maturin-extism builds. Tracked this down to [1]; the maintainers of `ring` don't intend to support linux sysroots that old. This PR updates the ARM Linux builds in particular to pull in a newer manylinux target. Tested on a [private fork](https://github.com/extism/dev-extism/actions/runs/6644247344) of the extism repo.

[1]: https://github.com/briansmith/ring/issues/1728


zshipko (approved):
Looks good! Thanks for digging into this

PR #552, opened 2023-10-27
feat: build static libextism

G4Vi:
* libextism is now built to crate-type `staticlib` as recommend by @zshipko
* updated `Makefile` to install and uninstall the static library.
* updated c example to have `static` target to link libextism statically
    * Compiler is no longer hardcoded to `clang`. Set `CC` to use a compiler other than the system compiler. Tested with `gcc` and `clang`.
* added `static-artifact` to actions release. Untested as it's only ran on push with tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,7 @@ jobs:
           - os: 'windows'
             target: 'x86_64-pc-windows-msvc'
             artifact: 'extism.dll'
-            static-artifact: 'libextism.lib'
+            static-artifact: 'extism.lib'
 
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -27,24 +27,31 @@ jobs:
...
             artifact: 'libextism.so'
+            static-artifact: 'libextism.a'
           - os: 'windows'
             target: 'x86_64-pc-windows-gnu'
             artifact: 'extism.dll'
+            static-artifact: 'libextism.a'
           - os: 'windows'
             target: 'x86_64-pc-windows-msvc'
             artifact: 'extism.dll'
+            static-artifact: 'libextism.lib'
```
zshipko, line 54: I think windows should be `extism.lib`
```suggestion
            static-artifact: 'extism.lib'
```

zshipko (commented):
Nice! I think the library name for windows is incorrect, but otherwise looks good. 

I will update the CLI to install the static library too

zshipko (approved):

PR #559, opened 2023-11-02
feat: pkg-config

G4Vi:
Adds `pkg-config` configs for dynamic and static builds of libextism. Example usage is shown in `libextism/Makefile`.

I added the `.skel` versions of the `.pc` files to non-MSVC releases. Ideally, the extism cli when installing the lib would set the PREFIX and install them.


zshipko (dismissed):
Works great, it's a major improvement to be able to locate the installation!

Typically I have seen these files with the `.pc.in` extension but since they're renamed before getting copied it doesn't really matter either way.

G4Vi:
> Typically I have seen these files with the `.pc.in` extension but since they're renamed before getting copied it doesn't really matter either way.

I could not think of the convention in the moment. Fixed!

zshipko (approved):

PR #571, opened 2023-11-06
feat: add x86_64-unknown-linux-musl target, specifying target to make

G4Vi:
The `x86_64-unknown-linux-musl` target does not build dynamic libraries for now:
```
 warning: dropping unsupported crate type `cdylib` for target `x86_64-unknown-linux-musl`
```

A target may now be optionally specified to the root  `Makefile`, example:
`make RUST_TARGET=x86_64-unknown-linux-musl && sudo make RUST_TARGET=x86_64-unknown-linux-musl install`

The new musl targets may be used to create entirely static binaries with libextism. See the new `musl-static` target in `libextism/Makefile`


zshipko (approved):
Thanks! 

I think the CLI is almost set up to handle this too, I need to take a look but I think we can just add a `libc` flag to specify which libc implementation/compiler to use.

G4Vi:
> I think the CLI is almost set up to handle this too, I need to take a look but I think we can just add a `libc` flag to specify which libc implementation/compiler to use.

Sounds good, maybe  `lib versions` should list which triplets are available for the current system.

PR #574, opened 2023-11-08
feat: add add_subdirectory compatible cmake config

G4Vi:
This enables cmake in-tree builds with libextism such as with the cpp-sdk: https://github.com/extism/cpp-sdk/pull/7

### DIFF
--- a/libextism/CMakeLists.txt
+++ b/libextism/CMakeLists.txt
@@ -1,0 +1,15 @@
+project(extism)
+cmake_minimum_required(VERSION 3.22)
+include(FetchContent)
+
+FetchContent_Declare(
+    Corrosion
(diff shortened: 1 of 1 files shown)


zshipko (approved):
Nice, tested and it works! I keep going back and forth on CMakeLists.txt in the root of the project vs in `libextism` but I think it makes sense like this since the documentation and everything is in that directory.

G4Vi:
> Nice, tested and it works! I keep going back and forth on CMakeLists.txt in the root of the project vs in `libextism` but I think it makes sense like this since the documentation and everything is in that directory.

It's worth considering. I avoided it as the root directory has a lot already and didn't want to create confusion with the root `Makefile`.

PR #578, opened 2023-11-10
refactor!: use `tracing` crate for logging, add `extism_log_callback` function

zshipko:
- Uses `tracing` instead of `log` crate
- Uses `tracing-subscriber` instead of `fern`
  - This allows us to automatically capture `log` events using `tracing-subscriber`
- Breaking: Makes `extism::set_log_file` private and only used through the C API, Rust users should use `tracing-subscriber` to determine which filters/levels to log.
- Adds `extism::set_log_callback` function to set a callback that can be used for custom logging from Rust. 
- Adds `bool extism_log_custom(const char *level)` and `extism_log_drain(void (*fn)(const char *s, size_t length)` to the C API to enable custom sinks in other SDKs


zshipko:
@mhmd-azeez I just pushed a commit to add `extism_log_callback`, which should allow users to handle logging themselves. If it works for you I will merge this.

zshipko:
Alright, just pushed a change to use `extism_log_custom(log_level)` and `extism_log_drain(callback)` instead of the single `extism_log_callback(callback, level)` function

zshipko:
@mhmd-azeez merging this now, but happy to make any adjustments if you end up running into any issues.

PR #583, opened 2023-11-14
cleanup: remove old SDKs

zshipko:
Removes all SDKs but the Rust SDK/runtime and the C SDK, which is automatically generated from the Rust crate.


.github/workflows/release-python.yaml
```diff
@@ -1,41 +0,0 @@
-name: Release Python SDK
```
chrisdickinson, line 1: Ah, we'll want to keep this workflow around – it's what implements the maturin `extism_sys` publish!

zshipko: thanks for taking a look, I will un-remove that one!

chrisdickinson (dismissed):
:tada: This looks great. You might already be thinking along these lines, but it might be worth pushing this to `extism/dev-extism` and running a fake release process there to shake out any GitHub actions bugs.

zshipko:
Great idea! I will make sure to test it on dev-extism first

zshipko:
Alright, just tested this `dev-extism` and the development release built successfully, so I'm going to merge it here! If I accidentally deleted anything we still need we can pull it from the git history.

PR #590, opened 2023-11-16
ci/cd: add building remanining rust tier 1 targets (i686 support)

G4Vi:
@tgraupmann mentioned a need for 32 bit x86 targets in discord.

I was surprised to learn they have tier 1 support with Rust: https://doc.rust-lang.org/rustc/platform-support.html#tier-1-with-host-tools Unfortunately, I don't believe wasmtime has x86 JIT support yet so these targets are probably slower.


G4Vi:
These targets actually won't work for now, wasmtime doesn't have interp mode: https://docs.wasmtime.dev/stability-platform-support.html#interpreter-support

Supporting these targets likely requires adding a new wasm engine to the runtime.

PR #596, opened 2023-11-17
feat: add support for wasmtime precompiled modules

zshipko:
- Adds `extism::compile` to AOT compile a wasm module from memory
- Adds `extism::Plugin::compile` to compile a Plugin
- Updates manifest loading to support pre-compiled modules
- Adds `create_plugin_cwasm` benchmark
- Adds `extism-compile` executable to compile a module from the CLI
- Doesn't handle any kind of automatic compilation or caching, we could experiment with building a caching layer on top of Extism in a new library

## Potential issues

Precompiled modules have to be compiled with the same settings used by the wasmtime runtime at execution time. An extism plugin can also be precompiled using wasmtime:

```
wasmtime compile -o wasm/code.cwasm wasm/code.wasm -W epoch-interruption -W tail-call -W function-references
```

But if `EXTISM_DEBUG` is enabled then `-D debug-info` will also need to be set. `extism-compile` can compile a module with these configurations set from the environment but there is not guarantee they will be compiled with the same options as they eventually are executed with.

## Performance
Loading a pre-compiled module shows some pretty significant speedups:

```
create/create_plugin    time:   [21.539 ms 23.307 ms 25.177 ms]                                 
                        change: [-42.871% -38.336% -33.149%] (p = 0.00 < 0.20)
                        Change within noise threshold.

create/create_plugin_cwasm                                                                             
                        time:   [3.5391 ms 3.6128 ms 3.6840 ms]
                        change: [-2.3246% +0.2421% +3.0526%] (p = 0.86 > 0.20)
                        No change in performance detected.

```


chrisdickinson:
This awesome to see! It does raise a question for me, though: given that one of our goals is to unify capabilities across different runtimes, how do we navigate this with runtimes on platforms that don't support pre-compilation (or require a different kind of precompilation?)

(Something that just came to mind: what if we treated compiled Wasm as a `@compiled` custom section inside an existing Wasm file? That is, `extism-compile` would take wasm as input and write a new wasm file with the original instructions plus a custom section containing flags + arch + a crc. Then `extism` proper could inspect incoming Wasm custom sections & optionally use the compiled Wasm if available & well-formed. Would that be too slow?)

zshipko:
That's a neat idea, I left this pretty minimal on purpose but am open to ways like that of improving the usability before merging. I can explore a little more next week!

johnpyp:
This looks great, and those load time performances are very very promising. I'm curious as well how this scales with increasing complexity of the WASM modules. If you had one that took ~100ms to load before, does it now take ~15ms to load with AOT, or does it take closer to 3ms?

runtime/src/manifest.rs
```diff
@@ -57,6 +57,15 @@ fn check_hash(hash: &Option<String>, data: &[u8]) -> Result<(), Error> {
 
 const WASM: &[u8] = include_bytes!("extism-runtime.wasm");
 
+fn new_module(engine: &Engine, buf: &[u8]) -> Result<Module, Error> {
+    let has_magic = buf.len() >= 4 && buf[0..4] == WASM_MAGIC;
+    if has_magic {
+        Ok(Module::new(engine, buf)?)
+    } else {
+        Ok(unsafe { Module::deserialize(engine, buf)? })
```
johnpyp, line 65: Could be potentially further sped up by leveraging https://docs.wasmtime.dev/api/wasmtime/struct.Module.html#method.from_trusted_file

It implicitly checks the magic bytes as well and has this same if/else. The major difference is it mmaps the file and some other stuff, reportedly making it faster.

bhelx:
> This awesome to see! It does raise a question for me, though: given that one of our goals is to unify capabilities across different runtimes, how do we navigate this with runtimes on platforms that don't support pre-compilation (or require a different kind of precompilation?)

Good question. We do also want to note that the pre-compiled binaries are not portable in anyway. you still want your source of truth, distributable code to be the wasm module.

zshipko:
@johnpyp it looks like it could scale pretty well, I tried again with a 5MB module (instead of ~1MB) and here are the results:

```
create/create_plugin    time:   [35.239 ms 36.560 ms 37.780 ms]                                 
                        change: [+48.073% +53.735% +58.927%] (p = 0.00 < 0.20)
                        Change within noise threshold.

create/create_plugin_cwasm                                                                             
                        time:   [3.6108 ms 3.6749 ms 3.7387 ms]
                        change: [-3.6813% -1.1732% +1.3403%] (p = 0.36 > 0.20)
                        No change in performance detected.

```

The normal load time has increased but the precompiled version is just about the same as before. I will need to find a bigger module to test with but it seems promising.

runtime/src/manifest.rs
zshipko: I had considered that, but we already have the file in memory at this point, so reading the file a second time feels a bit wasteful? Maybe there is some additional refactoring that could make this possible.

mhmd-azeez:
> Overall I think there are bound to be some differences in capabilities across the SDKs, but figuring out how not to make things fail unexpectedly is important. It seems like wazero has some support for precompilation so it should be possible there as well but I'm not sure about JS

Did some quick checks:

- wazero has Compiled Modules: https://pkg.go.dev/github.com/tetratelabs/wazero#CompiledModule
- JS has `WebAssembly.compile` which gives you a `Module` that can then be instantiated using `WebAssembly.instantiate`

None of them can instantiate a `.cwasm` file as far as I can see

zshipko:
Thanks for looking into it. I think it's fair to say that compiled code is runtime specific. We can provide functions in each runtime to precompile a module only but should document that these compiled modules aren't portable across runtimes.

chrisdickinson:
As a heads up, we're tracking support for using `WebAssembly.Module` as part of extism/js-sdk#9; it's not quite the same as precompiled modules, though, in that `WebAssembly.compile` doesn't prebuild to a given machine architecture – it's more akin to [`Module::new`](https://docs.rs/wasmtime/latest/wasmtime/struct.Module.html#method.new).

@zshipko Apologies, we talked a little bit about this in our team meeting because of extism/go-sdk#40! We were mostly brainstorming, no binding decisions or anything. To summarize:

After thinking about it a bit more, I realized that I'm a little nervous about committing to `.cwasm` as a public API via the manifest. It feels like an internal implementation detail of Extism; a `cwasm` file might not be compatible cross-machine, cross-sdk, or across versions of Extism. That was pushing me towards a "fat binary" approach, where we could embed multiple precompiled modules in a single stream, falling back to the appropriate one for the machine/sdk/extism version. However, @bhelx brought up an excellent counterpoint: we probably don't want that native code leaving the user's machine, at least for the time being – distribution and trust get tricky.

In discussion, @G4Vi mentioned exposing precompilation as an Extism option: if enabled, Extism would create a filesystem-backed "prebuild" cache. On the first load of a plugin, Extism would compile and store the precompiled module[^1]. This felt a lot better to me than trying to gate this functionality on storing native binaries in custom sections. What do you think?

[^1]: maybe using something like [cacache](https://docs.rs/cacache/latest/cacache/)?

johnpyp:
Having that as a transparent option rather than a separate API sounds like a great solution and will make it much easier for consumers to integrate too.

zshipko:
@chrisdickinson I kind of had something similar in mind with this PR kind of being an implementation detail. The Rust SDK already supports wat files despite not being supported in the JS or Go SDKs. It seems like figuring some of the automatic caching out in this PR would be beneficial so I can close up some parts of the API but it would be nice to still expose a way to allow users to have some level of control over when caching occurs.

zshipko:
Just pushed up an initial (very unoptimized) filesystem-based cache implementation and the benchmarks have improved even more due to the ability to pre-compile the Extism kernel. Automatic compilation caching can now be enabled with `PluginBuilder::with_cache_dir` or the `EXTISM_CACHE_DIR` environment variable.

```
Benchmarking create/create_plugin_cached: Warming up for 3.0000 s
create/create_plugin_cached                                                                             
                        time:   [1.7382 ms 1.7495 ms 1.7607 ms]
                        change: [-0.4932% +0.3121% +1.0022%] (p = 0.44 > 0.20)
                        No change in performance detected.
```

Since this seems like a viable approach I am going to keep working on cleaning this up a bit over the next few days and will  update when I'm satisfied with the cache implementation.

johnpyp:
I "backported" this PR to v0.5.4 to see how it could perform in a real-world scenario with proto (only the core component of `precompile_or_get_cached`... and the results are amazing!

For context, [proto](https://moonrepo.dev/docs/proto) is a version manager ala nvm, asdf, rtx, etc., and it loads a wasm plugin every time a cli command is run to resolve where to get the correct version of the command it's resolving for. For things like the `node` binary, this is very important as it gets recursively called constantly during execution of various build scripts, install scripts, postinstall scripts, etc.

Calling `node --version` should only take about 14ms normally. Using proto as the shim resolver, before this change, we were getting ~100ms execution time, with almost 100% of that time consumed by the loading of the wasm plugin:
```
hyperfine -- './target/release/proto run node -- --version'
Benchmark 1: ./target/release/proto run node -- --version
  Time (mean ± σ):     100.3 ms ±   4.4 ms    [User: 649.7 ms, System: 85.9 ms]
  Range (min … max):    91.8 ms … 110.7 ms    26 runs
```
That's +86ms over just running the binary directly!

...and now, with a cache dir set and the using the v0.5.4 backport, we're getting **~15ms execution time**! Only **+1ms** over directly running the binary! Outstanding:
```
hyperfine -- 'EXTISM_CACHE_DIR=./cache ./target/release/proto run node -- --version'
Benchmark 1: EXTISM_CACHE_DIR=./cache ./target/release/proto run node -- --version
  Time (mean ± σ):      15.4 ms ±   0.6 ms    [User: 13.6 ms, System: 18.7 ms]
  Range (min … max):    14.1 ms …  18.4 ms    163 runs
  ```
  
  ----
  For reference, here's the time to run `/usr/bin/node --version` on my system:
  ```
  hyperfine -- '/usr/bin/node --version'
Benchmark 1: /usr/bin/node --version
  Time (mean ± σ):      14.1 ms ±   2.4 ms    [User: 10.0 ms, System: 4.0 ms]
  Range (min … max):    11.1 ms …  22.2 ms    159 runs
```

Using the proto-provided node binary directly is somehow significantly faster than `/usr/bin/node` despite being basically the same version... maybe this is some permission thing not sure.

```
hyperfine --warmup 10 -- '~/.proto/tools/node/21.1.0/bin/node --version'
Benchmark 1: ~/.proto/tools/node/21.1.0/bin/node --version
  Time (mean ± σ):       2.2 ms ±   0.6 ms    [User: 0.8 ms, System: 1.6 ms]
  Range (min … max):     1.5 ms …   5.2 ms    546 runs
  ```

**Comment by mtb0x1** [comment]
The current proposal seems to be very advanced already 👍, I have two suggestions : (maybe a bit much :D ) 
- Cache invalidation (GC for cache) : keep track of cache usage and invalidate unused entries (reduce storage and memory usage during runtime)
- For JS (browser version) : risk of cache poisoning ? I don't see a viable cache mechanism other than storing the compiled blob into `indexeddb`, `localstorage` doesn't support more than 5MB per domain last time I checked.

**Comment by zshipko** [comment]
Cache invalidation is definitely something I would like to have some level of support for in this PR as long as it doesn't complicate things too much but browser support is something we'll have to investigate since I'm not certain there is an equivalent way to serialize compiled modules.

**Comment by mtb0x1** [comment]
> Cache invalidation is definitely something I would like to have some level of support for in this PR as long as it doesn't complicate things too much.

Maybe a first step would be to make sure when a plugin is unloaded/canceled its respective cache is remove and also expose a function that removes a cache entry using plugin id/name and let the user use when needed. 
The advanced stuff can be be implemented in different version.

**Comment by bhelx** [comment]
> Cache invalidation is definitely something I would like to have some level of support for in this PR as long as it doesn't complicate things too much but browser support is something we'll have to investigate since I'm not certain there is an equivalent way to serialize compiled modules.

> For JS (browser version) : risk of cache poisoning ? I don't see a viable cache mechanism other than storing the compiled blob into indexeddb, localstorage doesn't support more than 5MB per domain last time I checked.

Apparently there is a way to "compile" a module in the web APIs. But I assume it's not actually compiling it to machine code. I find it highly doubtful that you can stick some native code somewhere into a JS API and it will get executed... Some more research should be done there.

I'd personally be concerned about baking anything into Extism right now about caching or storing the compiled modules. I'd be more open to it if we discuss the details, but it seems kind of application specific.

> Maybe a first step would be to make sure when a plugin is unloaded/canceled its respective cache is remove and also expose a function that removes a cache entry using plugin id/name and let the user use it when needed.
The advanced stuff can be be implemented in different version.

Something to keep in mind is that using an AOT compiled module, in practice, should not be that different than loading the plugin and calling it at least once. The AOT module should be usable for the case where you need load the module from cold start quickly. Once you have the module loaded and the JIT has kicked in, the performance against the AOT module should not be noticeable.

**Comment by bhelx** [comment]
I'll follow up with Zach about my concerns. I may just not be understanding something.

**Comment by zshipko** [comment]
I am also not super interested in providing a full-featured cache, but I think giving the user an option to cache compiled modules to a particular directory is flexible enough for a lot of applications. I also think we should expose ways for users to handle their own caching when they need something more advanced. Also, by default this doesn't enable any caching, it only gets activated when a user specifies a cache directory.

Let's talk about it Monday, I am also hesitant to build out too much caching functionality, but I think the performance benefits are too good to ignore.

**Comment by chrisdickinson** [comment]
@mtb0x1:

> For JS (browser version) : risk of cache poisoning ? I don't see a viable cache mechanism other than storing the compiled blob into indexeddb, localstorage doesn't support more than 5MB per domain last time I checked.

Since there's no Web platform equivalent for Wasm modules that have been compiled to the target x86 or arm architecture ahead of time, this precompile caching feature seems likely to be limited to Extism SDKs running on top of Wasmtime (and eventually, Wazero.) Luckily, the Wasm implementations hosted by V8 and Spidermonkey have good support for fast, tiered JIT compilation (& [internal caching](https://v8.dev/blog/wasm-code-caching)) so plugin initialization time _should_ be less of an issue in the js-sdk. (I'm tempted to start thinking of this PR as "polyfilling V8's fast Wasm startup time")

@zshipko: If you're still interested in using cacache I can see about making a PR to make it possible just to use the sync APIs. Cacache implements the npm CLI's cache semantics in Rust, so it might be worth making it usable for our purposes, if only to offload some of the details of caching. (If not, that's totally cool, too!)

Also, apologies, I am _required_ to nerd out about this: if we _do_ go down the route of writing our own caching, there's a [neat trick](https://stackoverflow.com/a/18732276) we can borrow from git's loose object store to put a limit on the number of objects we're storing in this cache.

**Comment by zshipko** [comment]
> @zshipko: If you're still interested in using cacache I can see about making a PR to make it possible just to use the sync APIs. Cacache implements the npm CLI's cache semantics in Rust, so it might be worth making it usable for our purposes, if only to offload some of the details of caching. (If not, that's totally cool, too!)

It depends on what features we decide on, definitely worth discussing a bit more but the simple implementation using `Module::from_trusted_file` is pretty acceptable for now. For the most part, I'm still seeing this PR as adding very basic functionality while opening things up more for users to build upon that for their particular caching needs. 

Another option might be to use the built-in [wasmtime caching](https://docs.wasmtime.dev/cli-cache.html), the only real issue I ran into with that is that it's driven by a config file on disk, which could be fine if we document it well. I still need to explore this a bit more.

**Comment by zshipko** [comment]
It looks like using the wasmtime cache gets us most of the way there: 

```
create/create_plugin    time:   [2.9079 ms 2.9139 ms 2.9200 ms]                                  
                        change: [+3.2677% +3.6399% +3.9766%] (p = 0.00 < 0.20)
                        Change within noise threshold.

Benchmarking create/create_plugin_cwasm: Warming up for 3.0000 s
create/create_plugin_cwasm                                                                             
                        time:   [1.2195 ms 1.2279 ms 1.2368 ms]
                        change: [-1.7581% -0.7390% +0.2632%] (p = 0.17 < 0.20)
                        Change within noise threshold.
```

I still need to get a handle on how exactly it works and what options can be configured, but it seems like using the wasmtime cache would get us pretty close to the basic cache in this PR, but with some additional features (eviction, compression, ...) - my branch using wasmtime caching is here: https://github.com/extism/extism/tree/wasmtime-caching - this would reduce our maintenance burden a good amount but is a little less flexible.

johnpyp:
Once this is ready, can it be backported to v0.5.x as well? It would be great to use this in proto as soon as possible, especially since the integration is so easy :)

zshipko:
Closing in favor of https://github.com/extism/extism/pull/605

PR #597, opened 2023-11-19
Implement convenience traits for encoded types

george-hopkins:
Implement `From<T>` and `Debug` for `Base64<T>`, `Json<T>`, `Msgpack<T>` and `Protobuf<T>`


zshipko (approved):
Thanks! This is great as-is but I wonder if it makes sense to implement `From<Json<T>> for T` also?

george-hopkins:
Unfortunately, we can't implement `From<Json<T>>` on foreign types. We could implement `Into<T> for Json<T>` but this would prevent users from implementing `From<Json<T>>` themselves.

PR #603, opened 2023-11-21
release: 1.0.0-rc3

bhelx:


runtime/Cargo.toml
```diff
@@ -21,7 +21,7 @@ tracing-subscriber = {version = "0.3", features = ["std", "env-filter", "fmt"]}
 url = "2"
 glob = "0.3"
 ureq = {version = "2.5", optional=true}
-extism-manifest = { version = "1.0.0-rc2", path = "../manifest" }
+extism-manifest = { version = "1.0.0-rc3", path = "../manifest" }
```
chrisdickinson, line 24: Note for the future: I wonder if we can make this a `workspace = true` dep and manage all the versions from there?

chrisdickinson (approved):
LGTM!

runtime/Cargo.toml
bhelx: Yeah there is a little detangling that could be done here. And we should loosen these versions after 1.0

PR #604, opened 2023-11-22
build: drive crate versions from workspace; drive workspace version from ci

chrisdickinson:
This is an attempt to sand down [a sharp edge](https://github.com/extism/extism/actions/runs/6949120346/job/18906546065#step:4:476) around releases – keeping the `Cargo.toml` versions in-sync with the release tag, & the versions of the workspace crates aligned with one another.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 on:
   workflow_dispatch:
   push:
-    branches: [ main ]
+    branches: [ main, "v*" ]
     tags:
       - 'v*'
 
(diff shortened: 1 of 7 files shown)


zshipko (dismissed):
This is great! Keeping those versions in order has been a total pain. I'm curious if this changes anything when using extism as a git dependency? I can't really think of anything.

chrisdickinson:
Thanks! I don't _think_ this should change anything about using extism as a git dep, though folks might see the ugly `v0.0.0-replaced-by-ci` appear in output (e.g. `Compiling extism v0.0.0+replaced-by-ci (https://github.com/extism/dev-extism?tag=v0.0.9999#206c452d)`.) I was able to use it in a pretty straightforward way.

I'll merge this Monday – I'd like to flag that I haven't taken this through the _complete_ release process yet (to release to cargo/pypi), but I've built the main branch and a tag of dev-extism with it successfully. After I merge I'll roll an rc release, just to make sure nobody trips over a rake I've left in the yard 😅

PR #612, opened 2023-11-28
cleanup: simplify main module resolution

zshipko:
This PR simplifies the resolution of the `main` module when multiple modules are provided. Before we would try to look at the path/URL when the wasm was coming from disk or via HTTP, now any module missing a name will be used as `main`. This is much nicer and more consistent since this is what was being done when no filename was available (i.e. raw data modules). It also make sense because non-main modules will need to be named for the functions to be linked correctly.

### DIFF
--- a/runtime/src/manifest.rs
+++ b/runtime/src/manifest.rs
@@ -75,7 +75,6 @@ fn to_module(engine: &Engine, wasm: &extism_manifest::Wasm) -> Result<(String, M
             meta,
         } => {
             // Get the file name
-            let file_name = url.split('/').last().unwrap_or_default();
             let name = match &meta.name {
                 Some(name) => name.as_str(),
                 None => meta.name.as_deref().unwrap_or(MAIN_KEY),
@@ -178,8 +177,12 @@ pub(crate) fn modules(
         return Ok(());
     }
 
-    for f in &manifest.wasm {
-        let (name, m) = to_module(engine, f)?;
+    for (i, f) in manifest.wasm.iter().enumerate() {
+        let (mut name, m) = to_module(engine, f)?;
+        // Rename the last module to `main` if no main is defined already
+        if i == manifest.wasm.len() - 1 && !modules.contains_key(MAIN_KEY) {
+            name = MAIN_KEY.to_string();
+        }
         if modules.contains_key(&name) {
             anyhow::bail!("Duplicate module name found in Extism manifest: {name}");
         }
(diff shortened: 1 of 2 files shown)


nilslice:
does this impact anything in the go or js SDKs? cc/ @mhmd-azeez

chrisdickinson:
@nilslice While the JS SDK supports loading multiple modules, it doesn't currently implement linking (https://github.com/extism/js-sdk/issues/29). (I'd like to get that done as part of the post-beta/pre-release polish, though.)

nilslice:
@chrisdickinson ok great, thanks! still probably worth a check on the Go SDK, but if I recall there is some existing special checking done to handle this.. just want to be sure we are consistent

runtime/src/manifest.rs
```diff
@@ -45,10 +45,7 @@ fn to_module(engine: &Engine, wasm: &extism_manifest::Wasm) -> Result<(String, M
 
             // Figure out a good name for the file
             let name = match &meta.name {
-                None => {
-                    let name = path.with_extension("");
-                    name.file_name().unwrap().to_string_lossy().to_string()
-                }
+                None => meta.name.as_deref().unwrap_or(MAIN_KEY).to_string(),
```
chrisdickinson, line 48: Do we want to keep track of the names we've seen so we can error on duplicates? (It's possible we'll use the default `MAIN_KEY` multiple times if there are multiple unnamed modules.)

zshipko:
Good point, I can take a look at the go-sdk too and if I can't figure it out will just open an issue.

runtime/src/plugin.rs
```diff
@@ -222,6 +223,11 @@ impl Plugin {
 
         let engine = Engine::new(&config)?;
         let (manifest, modules) = manifest::load(&engine, wasm)?;
+        if modules.len() <= 1 {
+            anyhow::bail!("No wasm modules provided");
```
chrisdickinson, line 227: Has the kernel been injected by this point? (I dimly recall that we were injecting it while loading the manifest)

runtime/src/manifest.rs
zshipko: Thanks, just added a check for duplicate module names!

chrisdickinson (approved):
LGTM! This is a nice refinement. The only open question I had was whether the kernel counted towards our minimum module count check.

runtime/src/plugin.rs
zshipko: Yeah, `manifest::load` returns the user modules in addition to the kernel module

chrisdickinson: _slaps forehead_ Aaah yeah, gotcha – I was misreading the comparison, sorry. You're already taking that extra module into account! Please disregard 😅

zshipko:
Merging this now and opened a PR to update the go-sdk: https://github.com/extism/go-sdk/pull/42

PR #622, opened 2023-12-04
fix: improve handling of 0 length blocks

### LINKED ISSUE #620: Error when passing an empty String from Plugin to Runtime via a host function or a logging macro
First of all, thank you so much for Extism. Settled with it for a personal project and every time I'm using it I'm like: "oh shit, it's amazing :exploding_head:".

I'm using `extism@1.0.0-rc6` and `extism_pdk@1.0.0-rc1`, both in stable Rust on Linux. I've encountered the following error while iterating over the lines of a file and "sending" the lines back to host via a host function. I can't share the code right now, but I think I've managed to pin down the problem: **the error occurs when an empty `String` is "sent" from a plugin to the runtime, but only when "sent" via a host function or one of the `log` macros**. Here's a minimal example of the error, I hope it's reproducible at your end.

**First, the plugin code**

```rust
use extism_pdk::*;

#[plugin_fn]
pub fn test() -> FnResult<String> {
    let empty = String::new();
    info!("{}", empty);
    Ok(empty)
}
```

**Then the runtime code which calls the plugin**

```rust
use extism::{Manifest, PluginBuilder, Wasm};

fn main() {
    // Load and build plugin
    let url = Wasm::file("../my-plugin/target/wasm32-wasi/debug/my_plugin.wasm");
    let manifest = Manifest::new([url]);
    let mut plugin = PluginBuilder::new(manifest)
        .with_wasi(true)
        .build()
        .unwrap();
    // Call test method, this does not work
    let p = plugin.call::<(), String>("test", ()).unwrap();

    println!("{}", p);
}
```

**What happens**

I get the following error message:

```
thread 'main' panicked at src/main.rs:12:51:
called `Result::unwrap()` on an `Err` value: error while executing at wasm backtrace:
    0: 0x201d - extism_pdk::memory::Memory::log::h0d2eb2e83c17d446
                    at /home/robin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/extism-pdk-1.0.0-rc1/src/memory.rs:113:35
    1: 0x16d2 - my_plugin::test::inner::h18149dba2a840a95
                    at /home/robin/test/my-plugin/src/lib.rs:6:5
    2:  0xfd4 - test
                    at /home/robin/test/my-plugin/src/lib.rs:3:1
    3: 0x11d1b - <unknown>!test.command_export

Caused by:
    invalid handle offset: 0
```

When I check at the plugin side if the `String` is empty before sending it and don't send it if it is empty, everything works as expected. Also when I return an empty `String` as a plugin function return value, it works.

I would expect that it's possible to "send" back an empty `String`, or that I'm at least warned somehow when it's not possible for technical reasons. I'm too inexperienced to provide a fix, but feel free to reach out if you need any more info / testing.

Best regards,
rob

zshipko:
Fixes #620 

- Returns `MemoryHandle::null()` instead of returning an error when the offset is 0


nilslice (approved):
Thanks for the quick turnaround!

PR #625, opened 2023-12-06
build: add c++ compat to extism.h

G4Vi:
Before `extern "C"` was needed to include the header in C++ code. Now that's included inside, so it's easier and cleaner to include in c++ projects like the `cpp-sdk`.

### DIFF
--- a/runtime/extism.h
+++ b/runtime/extism.h
@@ -3,10 +3,6 @@
 #include <stdint.h>
 #include <stdbool.h>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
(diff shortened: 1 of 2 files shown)


runtime/extism.h
```diff
@@ -3,6 +3,10 @@
 #include <stdint.h>
 #include <stdbool.h>
 
+#ifdef __cplusplus
```
zshipko, line 6: This file is generated by cbindgen, I think you can do this by calling [Builder::with_cpp_compat](https://docs.rs/cbindgen/latest/cbindgen/struct.Builder.html#method.with_cpp_compat) in `build.rs`: https://github.com/extism/extism/blob/main/runtime/build.rs

zshipko (approved):

PR #626, opened 2023-12-08
feat: Add `extism_convert::Raw` to allow direct encoding using bytemuck

zshipko:
- Adds `extism_convert::Raw` to encode certain types using their direct memory representations using https://github.com/Lokathor/bytemuck
- Only enabled on little-endian targets to prevent memory mismatch issues 
- Allows for certain types of structs to be encoded using their in-memory representation
- Makes passing structs between Rust and C easier since none of the other encodings are available in C by default.

## Usage
After making a bytemuck-compatible struct:
```rust
use bytemuck::{Zeroable, Pod};

#[derive(Debug, Clone, Copy, PartialEq, Pod, Zeroable)]
#[repr(C)]
struct Point {
  x: i32,
  y: i32,
}
```

It can be used in `call`:

```rust
let input = Point { x: 100, y: 50 };
let Raw(pt): Raw<Point> = plugin.call("transform", Raw(&input))?;
```

### DIFF
--- a/convert/Cargo.toml
+++ b/convert/Cargo.toml
@@ -24,3 +24,4 @@ serde = { version = "1.0.186", features = ["derive"] }
 default = ["msgpack", "protobuf"]
 msgpack = ["rmp-serde"]
 protobuf = ["prost"]
+raw = []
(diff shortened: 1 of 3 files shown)


nilslice (approved):
great idea!

PR #627, opened 2023-12-12
fix(kernel): improve performance after large allocations, add `extism_plugin_reset` to give users more control when dealing with large allocations

zshipko:
See https://github.com/extism/cpp-sdk/issues/15

- Limits a call to memset in the kernel to the size of the current memory offset instead of the total size of memory.
- Adds `extism_plugin_reset` to the C API and `extism::Plugin::reset` to Rust


G4Vi (approved):
LGTM! This makes performance much more predictable. I tested out `extism_plugin_reset` and that works as well.

Issue #141, opened 2022-12-06
Improve windows support


nilslice:
Need to work on:
- windows CI
- WASI tests (see issue re [different platform behavior](https://discord.com/channels/1011124058408112148/1011124061100843020/1049509012284129423))
- builds & packaging DLL (possible overlap with #119)

brad-jones:
What is the current state of Windows support? I noticed there are windows release artifacts. I guess I can only download them & try them out.

Super cool project :)

nilslice:
Thanks, @brad-jones! I'd say it's still relatively untested as far as the install process goes. However, it's been pretty well tested in .NET on windows which we know works well. 

If you have access to WSL/2 environment, that's probably the best bet for now, until we get some more windows help. 

Maybe you'd be interested in helping us test and see where the shortcomings are?

bhelx:
One good tip we got here was that we make heavy use of Makefiles and unix utilities in our builds. We should be using a cross platform tool like cmake.

bhelx:
ARM windows support would also be good

benwiley4000:
Hi there, I saw there's not currently Windows support for the JavaScript PDK. Are there known blockers that need to be addressed, or has it just not been tested? I'm wondering if it has a good shot of working in Windows Subsystem for Linux.

nilslice:
WSL should work! 

I think the lack of Windows support is both a combination of few experienced Windows users on the team, and some missing install scripts or equivalent instructions for things like the WASI SDK needed for some of the internals.

benwiley4000:
Awesome thanks for the info.

Ben

bhelx:
Yes, I don't really see any major blockers as to why it wouldn't work. The main tools are rust and wasi-sdk (which both work on windows). Perhaps the make build system wouldn't work but there is not much in the Makefile. I didn't do it because i didn't have a reliable way to test it.

bhelx:
Feel free to file an issue on the js-pdk repo and we can track it. I can help if you can test. I do think we have other windows users in the community who may be able to help if we need.

bhelx:
What is missing here? Could we perhaps file specific issues for follow ups on windows support?

PR #632, opened 2023-12-13
ci: Release extism.dll.lib and extism.dll.a in the artifacts for Windows users

mhmd-azeez:
`extism.dll.lib` includes the imports for `extism.dll` and is necessary for MSVC if the user wants to link dynamically to extism.dll. This file is already generated by the build process, we just need to bundle it in the artifacts

Related to #141


mhmd-azeez:
The branch name is causing problems for the CI

Issue #584, opened 2023-11-15
Improve MSVC support


G4Vi:
The msvc releases do not include an import library (`.lib`) for `extism.dll`, only the static library `.lib`. The import library makes it possible to "link" to a DLL at build time instead of needing to `LoadLibrary` and `GetProcAddress` to use extism functions.

The libextism cmake file hasn't been tested on Windows and probably should specify the needed dependencies to link the static version.

Related to https://github.com/extism/extism/issues/141

mhmd-azeez:
> add import .lib to releases
This will be handled by #633

mhmd-azeez:
@G4Vi 
> test and fix libextism cmake file with msvc

is this still relevant?

G4Vi:
@mhmd-azeez 

That is still relevant. However, another item to consider is vcpkg support. vcpkg support should enable users to build a project using libextism on all systems with the same cmake file.

There are two goals I'd like to accomplish:

1. Have a familiar / standard development experience using libextism on Windows  (easy to use with visual studio c++)
2. Support cross-platform projects, use same repo and project files with minimal duplication, to build a project using libextism for Windows, Mac, Linux. (cmake, also cmake in-tree builds)

Goal 1 is more important than goal 2, but both are important. I trust your judgement on what next to proceed on.

I have revised this ticket with items I believe related to goals 1 and 2. Please let me know if you think we should go in a different direction to accomplish those goals.

Info on vcpkg:
* https://learn.microsoft.com/en-us/vcpkg/get_started/overview
* https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-packaging?pivots=shell-cmd

mhmd-azeez:
Seems like vcpkg doesn't support rust libraries yet, and I couldn't find a workaround:
- https://github.com/microsoft/vcpkg/pull/30251
- https://github.com/microsoft/vcpkg/issues/20619

But, there is some hope:
https://github.com/microsoft/vcpkg/pull/30251#issuecomment-1639787175

We can revisit it later on after Rust support is added

PR #633, opened 2023-12-13
ci: release extism.dll.lib and extism.dll.a

mhmd-azeez:
Related to #141 and #584 
Follow up of #632

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,18 +85,18 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      # - name: Set version
-      #   shell: bash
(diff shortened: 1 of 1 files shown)


mhmd-azeez:
Here is a sample project that shows this works:
https://github.com/mhmd-azeez/extism-cpp-demo

G4Vi (approved):
LGTM, thank you! At some point we should test building with mingw. It's good to have all the import libraries.

chrisdickinson (approved):
LGTM! Great work.

chrisdickinson (approved):
Github didn't clear my notifications with that last LGTM, so you get two LGTMs from me! 😅

PR #635, opened 2023-12-15
fix(kernel): length function should return 0 for invalid offsets

### LINKED ISSUE #634: extism_length returns garbage values when an ExtismPointer is not aligned with an alloc
```c
uint8_t outputConfig[12];
memset(outputConfig, 'A', 12);
ExtismPointer ep = extism_alloc_string((const char *)outputConfig, sizeof(outputConfig));
static char message[256];
snprintf(message, sizeof(message), "the lengths %u %u %u %u\n", ((uint32_t)extism_length(ep)), ((uint32_t)extism_length(ep + 1)), ((uint32_t)extism_length(ep + 2)), ((uint32_t)extism_length(ep + 3)));
extism_log(message, strlen(message), ExtismLogError);
```
`2023-12-14T23:26:30.583608Z ERROR extism::pdk: the lengths 12 1090519040 1094778880 1094795520`

Looking at `fn length` in `kernel/src/lib.rs`  it looks like it's supposed to return 0 when a block isn't found.

zshipko:
Fixes #634 

- Updates `extism_length` to walks the allocation list to determine valid offsets instead of assuming the provided offset is valid


G4Vi (approved):
Tested it out and it works great! Eventually, it may be worth exposing a block so iterating across the blocks isn't required to get the length, but I think this is a good idea for `extism_length` given the flexibility of an `ExtismPointer`.

PR #638, opened 2023-12-15
fix: give extism_log_drain's param a named type

chrisdickinson:
This fixes a test failure on the python-sdk [1]. (See also [2]). In particular, while maturin creates bindings for `extism_log_drain` on clang platforms, it seems that MSVC cannot generate the required binding information for `ffi.py`. This results in an `extism_sys` wheel whose shared object (in this case, a DLL) contains the `extism_log_drain`, but whose Python FFI bindings don't contain a definition for that function.

Naming the function pointer type parameter resolves the issue. What a strange issue!

[1]: https://github.com/extism/python-sdk/actions/runs/7172934060/job/19669775001#step:9:35
[2]: https://github.com/extism/python-sdk/pull/18#issuecomment-1850892835

### DIFF
--- a/runtime/extism.h
+++ b/runtime/extism.h
@@ -97,6 +97,11 @@ typedef void (*ExtismFunctionType)(ExtismCurrentPlugin *plugin,
                                    ExtismSize n_outputs,
                                    void *data);
 
+/**
+ * Log drain callback
+ */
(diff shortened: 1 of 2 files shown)


zshipko (approved):
Definitely a strange one!

PR #644, opened 2023-12-27
Toggle off default `clap` feature for `cbindgen`

CosmicHorrorDev:
The only feature for `cbindgen` is the `clap` feature for using it as a standalone binary and isn't needed when using it as a library


zshipko (approved):
Thanks!

PR #647, opened 2024-01-02
cleanup(kernel): add handle_length, restore faster length implementation

zshipko:
- Restores kernel `length` implementation to the version that avoids walking the allocation list
- Adds `handle_length` function, which is slower, but will return 0 if an offset isn't a valid handle


zshipko:
Still thinking about this one...

PR #648, opened 2024-01-03
feat(kernel): add extism_length_unsafe

### LINKED ISSUE #643: Consider reverting `length` safety improvement
Previously, I noticed the `length` function doesn't work on all `ExtismPointer`'s in https://github.com/extism/extism/issues/634 . After studying the issue more, I realized the issue is less an issue with the API, but more of a type safety issue in the pdks. In the C PDK, I've been working on `ExtismHandle`, a subset of `ExtismPointer` that always refers to the start of the data section of a memory block https://github.com/extism/c-pdk/pull/18 . 

Upon noticing several places `length` is called, for efficiency reasons I think we should revert https://github.com/extism/extism/commit/fd95729d8d668a8777a45bc779f7d8cd1c3559e7, possibly keeping the safe implementation under a new name, though I don't think that's necessary as long as the pdks differentiate the types.

zshipko:
- Adds `length_unsafe` function to the extism kernel, a more performant `length` for known-valid memory handles

After this is merged I will update go-sdk and js-sdk too.

Closes #643


G4Vi (approved):
LGTM, thank you!

PR #650, opened 2024-01-05
cleanup: improve wat detection

zshipko:
- Update to ensure `is_wat` is only true when the input is a valid string
- Update to allow leading whitespace in a wat file

### DIFF
--- a/runtime/src/manifest.rs
+++ b/runtime/src/manifest.rs
@@ -118,7 +118,8 @@ pub(crate) fn load(
         WasmInput::Data(data) => {
             let has_magic = data.len() >= 4 && data[0..4] == WASM_MAGIC;
             let is_wat = std::str::from_utf8(&data).is_ok_and(|data| {
-                data.trim_start().starts_with("(module") || data.trim_start().starts_with(";;")
+                let data = data.trim_start();
+                data.starts_with("(module") || data.starts_with(";;")
             });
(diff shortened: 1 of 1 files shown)


runtime/src/manifest.rs
```diff
@@ -117,10 +117,14 @@ pub(crate) fn load(
     match input {
         WasmInput::Data(data) => {
             let has_magic = data.len() >= 4 && data[0..4] == WASM_MAGIC;
-            let is_wat = data.starts_with(b"(module") || data.starts_with(b";;");
+            let s = std::str::from_utf8(&data);
+            let is_wat = s.is_ok_and(|data| {
+                let data = data.trim_start();
+                data.starts_with("(module") || data.starts_with(";;")
```
chrisdickinson, line 123: (Ooh, this could also start with `(;`)

zshipko: ah thanks! i will open a new PR and link here. would `(;` be the start of a comment inside the opening paren?

chrisdickinson: It's a block comment (there's an example [over here](https://github.com/dylibso/component-workbench/blob/main/00-core-module-greet.wat#L1-L6)!)

zshipko: also, would a paren followed by whitespace be valid? something like this:

```wat
(
module
```

PR #651, opened 2024-01-05
cleanup: improve wat detection (again)

zshipko:
Merged the other one too quickly, this PR is updated with some additional feedback from @chrisdickinson:

              (Ooh, this could also start with `(;`)
https://github.com/extism/extism/pull/650#discussion_r1443419249

It's also updated to accommodate for modules that start with an open paren followed by some whitespace. I doubt this covers all of the permitted syntax but it should be good enough.


chrisdickinson (approved):
LGTM!

PR #652, opened 2024-01-05
chore: kernel add Handle and Offset types to differentiate from Pointer

G4Vi:
This cosmetic change to the kernel adds two `u64` types for usage where `Pointer` was used and it is actually something else or is more restrictive (a pointer to the start of a data block (handle) or a relative offset). 

Instead of adding `Offset` to match `Length` I'd prefer to use `u64` directly when referring to those scalars, but not sure if the those types are preferred?

### DIFF
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -40,8 +40,6 @@ use core::sync::atomic::*;
 
 pub type Pointer = u64;
 pub type Handle = u64;
-pub type Length = u64;
-pub type Offset = u64;
 
 /// WebAssembly page size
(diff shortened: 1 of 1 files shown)


zshipko:
The only reason for `Pointer`/`Length` was to keep everything straight when I was initially working on it - I think we could use `u64` where it makes sense now.

G4Vi:
That makes sense. To avoid confusion with the pointer types I'd like to keep `Pointer` and `Handle` around.  I removed `Offset` and `Length`

zshipko (approved):

PR #653, opened 2024-01-05
feat: add rust-protobuf support to extism-convert

nilslice:
(code from @zshipko, thank you!)

### DIFF
--- a/convert/Cargo.toml
+++ b/convert/Cargo.toml
@@ -23,8 +23,6 @@ serde_json = "1.0.105"
 serde = { version = "1.0.186", features = ["derive"] }
 
 [features]
-default = ["msgpack", "protobuf", "raw"]
+default = ["msgpack", "prost", "raw"]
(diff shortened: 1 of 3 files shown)


dselans:
❤️

zshipko:
I think the CI failure is because when it's built with `--all-features` the two `Protobuf`  types conflict

I'm wondering if we should just break the current feature names an use `prost` to enable the `Prost` type/`prost` crate and `protobuf` to enable the `Protobuf` type/`protobuf` crate. It seems like long-term that makes the most sense, even if it's kind of weird for current users right now.

nilslice:
> I'm wondering if we should just break the current feature names an use prost to enable the Prost type/prost crate and protobuf to enable the Protobuf type/protobuf crate. It seems like long-term that makes the most sense, even if it's kind of weird for current users right now.

yea, I agree. I've updated to follow this plan.

zshipko (approved):

PR #659, opened 2024-01-17
fix(kernel): Fix calculation of handle offset when splitting re-used memory, add kernel test

Marton6:
Check out this branch locally, then

```
cd kernel
./test.sh
```

to run the test.

The test will fail, indicating an issue I encountered.


zshipko:
Thank you so much for this great reproduction, I really appreciate it! 

I think I have a fix here: https://github.com/extism/extism/pull/new/fix-kernel_memory_bug

The issue was when the re-use offsets were being calculated we weren't accounting for the size of the `MemoryBlock` structure so `MemoryBlock::status` was being overwritten for second half of the split memory.

zshipko:
I just verified my change works against CI - I am going to push my commit to this branch and merge it shortly.

Thanks again for your report!

zshipko:
Hm, that seems like a weird CI failure,  I'm going to open a separate PR to test against.

zshipko:
@Marton6 let me know if you continue to run into any issues when using the latest `main` branch - if everything is working we can make a new bugfix release

PR #667, opened 2024-01-27
feat(convert): add derive macros for To and FromBytes

### LINKED ISSUE #661: derive that forwards From/ToBytes to specific encoding
My idea was to add to `extism_convert` either derive macros for the different data formats, or a single `Encoding` derive macro, that specifies a specific one.

### In:
```rs
#[derive(ToJson, FromJson)]
// or
#[derive(ToEncoding, FromEncoding)]
#[encoding(Json)]
struct Struct { /* some fields */ }
```

### Out:
```rs
// ToJson
impl<'a> ToBytes<'a> for Struct {
  type Bytes = <Json<Self> as ToBytes<'a>>::Bytes;
  
  fn to_bytes(&self) -> Result<Self::Bytes, Error> {
    Json(self).to_bytes()
  }
}
// FromJson
impl FromBytesOwned for Struct {
  fn from_bytes_owned(date: &[u8]) -> Result<Self, Error> {
    Json::from_bytes_owned(data).map(|j| j.0)
  }
}
```

This allows using it directly as function arguments/return type, and it automatically has the correct encoding (e.g., when sharing code between host and plugin)

ModProg:
closes #661.

- [x] docs
- [x] tests
- [x] depend on `extism-convert/extism-pdk-path` feature in https://github.com/extism/rust-pdk https://github.com/extism/rust-pdk/pull/47


convert-macros/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+            }
+            Ok(FoundCrate::Itself) => parse_quote!(crate),
+            Err(_) => match () {
+                () if cfg!(feature = "extism-path") => parse_quote!(::extism::convert),
+                () if cfg!(feature = "extism-pdk-path") => parse_quote!(::extism_pdk),
+                _ => parse_quote!(::extism_convert),
+            },
+        },
+    }
+}
```
ModProg, line 29: This feature based fallback logic is likely redundant. I implemented it this way in a similar multicrate project but don't remember the issues this solved: https://github.com/khonsulabs/bonsaidb/blob/8c9f47f97754e7613a18a24c092cb48423778acf/crates/bonsaidb-macros/src/lib.rs#L35-L61

convert-macros/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+        parse_quote!(for<'__to_bytes_b> #encoding<&'__to_bytes_b Self>: #convert::ToBytes:<'__to_bytes_b>)
+    );
+    generics.params = iter::once(parse_quote!('__to_bytes_a))
+        .chain(generics.params)
+        .collect();
+    let (impl_generics, type_generics, where_clause) = generics.split_for_impl();
+    Ok(quote! {
+        impl #impl_generics #convert::ToBytes<'__to_bytes_a> for #ident #type_generics #where_clause
+        {
+            type Bytes = ::std::vec::Vec<u8>;
```
ModProg, line 64: I tried using the type of the encoding, i.e. `type Bytes = <Json<&Self> as ToBytes>::Bytes;` but was unable to figure out the correct lifetimes.

zshipko: For values encoded with serde, it's likely that `Self::Bytes` is already `Vec<u8>`, so I don't think that's a major issue. The `Raw` encoding might use `&[u8]`, but since this is a higher-level feature I think we can ignore the super high performance case.

convert-macros/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+use syn::{parse_quote, Attribute, DeriveInput, Path};
+
+/// Tries to resolve the path to `extism_convert` dynamically, falling back to feature flags when unsuccessful.
+fn convert_path() -> Path {
+    match crate_name("extism") {
+        Ok(FoundCrate::Name(name)) => {
+            let ident = format_ident!("{name}");
+            parse_quote!(::#ident::convert)
+        }
+        Ok(FoundCrate::Itself) => parse_quote!(crate::core),
```
zshipko, line 15: This line was giving me some trouble, but seems to work if I remove `core`:
```suggestion
        Ok(FoundCrate::Itself) => parse_quote!(crate),
```

zshipko (commented):
Looks great - I have a commit where I changed some of the tests to use this and it makes the developer experience very nice! Those changes are here: https://github.com/extism/extism/tree/test-convert-derive-macros

I'm not 100% clear how the right way to do the path resolution in this case, but I think your solution works pretty well - as long as the end user doesn't need to add any additional features when using the `extism` or `extism-pdk` crates.

I think we just need some more tests and documentation (like you mentioned in your issue) and it should be ready!

convert-macros/src/lib.rs
ModProg: oh sorry that was a copy paste :D

ModProg:
> as long as the end user doesn't need to add any additional features when using the `extism` or `extism-pdk` crates

Yeah, that is the idea. This would need to be extended for any “entry point” crates that are created in the future. But sadly, rust still doesn't have a `$crate` for proc macros.

zshipko:
Hey @ModProg - how's this going? If you're stuck on the documentation/testing I can help pick some of that up next week.

convert-macros/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[lib]
+proc-macro = true
+
+[features]
+extism-path = []
+extism-pdk-path = []
+
+[dependencies]
+# manyhow = "0.10.4"
```
ModProg, line 20:
```suggestion
```

convert/Cargo.toml
```diff
@@ -18,11 +18,15 @@ protobuf = { version = "3.2.0", optional = true }
 rmp-serde = { version = "1.1.2", optional = true }
 serde = "1.0.186"
 serde_json = "1.0.105"
+extism-convert-macros.workspace = true
 
 [dev-dependencies]
+# extism = { path = "../runtime" }
```
ModProg, line 24:
```suggestion
```

convert/src/lib.rs
```diff
@@ -5,6 +5,9 @@
 //! similar to [axum extractors](https://docs.rs/axum/latest/axum/extract/index.html#intro) - they are
 //! implemented as a tuple struct with a single field that is meant to be extracted using pattern matching.
 
+// Makes proc-macros able to resolve `::extism_convert` correctly
+extern crate self as extism_convert;
+
```
ModProg, line 10: These will only be relevant as soon as proc macros are used inside this crate. So not yet. But this will avoid any future confusion.

runtime/Cargo.toml
```diff
@@ -22,7 +22,7 @@ url = "2"
 glob = "0.3"
 ureq = {version = "2.5", optional=true}
 extism-manifest = { workspace = true }
-extism-convert = { workspace = true }
+extism-convert = { workspace = true, features = ["extism-path"]}
```
ModProg, line 25:
```suggestion
extism-convert = { workspace = true, features = ["extism-path"] }
```

runtime/src/lib.rs
```diff
@@ -1,8 +1,12 @@
+// Makes proc-macros able to resolve `::extism` correctly
+extern crate self as extism;
+
 pub(crate) use extism_convert::*;
 pub(crate) use std::collections::BTreeMap;
 use std::str::FromStr;
 pub(crate) use wasmtime::*;
 
+#[doc(inline)]
 pub use extism_convert as convert;
```
ModProg, line 10: This makes the `convert` crate show up as a module instead of a reexport. This makes the paths inside the examples correct.

convert/Cargo.toml
```diff
@@ -18,11 +18,16 @@ protobuf = { version = "3.2.0", optional = true }
 rmp-serde = { version = "1.1.2", optional = true }
 serde = "1.0.186"
 serde_json = "1.0.105"
+extism-convert-macros.workspace = true
 
 [dev-dependencies]
+# Only required for tests run from the workspace root, as that enables the `extism-path` feature.
+extism.workspace = true
```
ModProg, line 25: Alternatively, we could add a hidden `# mod extism {pub use extism_convert as convert;}` to the doc tests.

ModProg:
The last too commits feel a bit like hacks to make ci happy :D

zshipko:
> The last two commits feel a bit like hacks to make ci happy :D

Ah yeah the path resolution features seem to add some additional complexity around docs/testing but it seems worth it.

I think these changes looks great and appreciate the effort updating the docs! I will merge this and the PDK change later after I experiment with the changes a little more.

zshipko (approved):

PR #671, opened 2024-02-02
refactor!: use wasi preview2

zshipko:
- Breaking: No longer copies Extism config values into WASI environment variables, these should be accessed using the Extism config functions instead
- Requires wasmtime 17 or greater
- Allows WASI TCP/UDP requests to hosts listed in the `allowed_hosts` field of the manifest
- Adds support for wasi-preview2

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -9,8 +9,8 @@ repository.workspace = true
 version.workspace = true
 
 [dependencies]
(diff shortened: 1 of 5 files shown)


zshipko:
This is now pretty out of date - since we would need to upgrade wasmtime versions before moving to wasip2 I am going to close this in favor of https://github.com/extism/extism/pull/697 - we can circle back on the wasip2 support when it makes sense for us.

PR #685, opened 2024-02-22
fix: linker issue that depends on the ordering of the linked functions

zshipko:
It looks like when a module is added to the linker, all of its imports must already be present. This PR updates the linking process to take that into consideration and adds a test with a reproduction of the issue @chrisdickinson shared with me.


chrisdickinson (approved):
LGTM! Thanks for taking a look at this.

PR #692, opened 2024-03-06
feat: add ability to configure size of the Extism var store

zshipko:
- Adds `memory.max_var_bytes` to the manifest to limit the number of bytes allowed to be stored in Extism vars - if `max_var_bytes` is set to 0 then vars are disabled.
- Adds some builder functions to `MemoryOptions` struct
- Sets the default var store size to 1mb
- Includes a test to make sure `var_set` returns an error when the limit is reached

### DIFF
--- a/runtime/src/pdk.rs
+++ b/runtime/src/pdk.rs
@@ -104,8 +104,8 @@ pub(crate) fn var_set(
 
     let voffset = args!(input, 1, i64) as u64;
 
-    // If the store is larger than 100MB then stop adding things
-    if size > 1024 * 1024 * 100 && voffset != 0 {
+    // If the store is larger than the configured size, or 1mb by default, then stop adding things
+    if size > data.manifest.memory.max_var_bytes.unwrap_or(1024 * 1024) as usize && voffset != 0 {
         return Err(Error::msg("Variable store is full"));
     }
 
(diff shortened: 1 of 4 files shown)


runtime/src/pdk.rs
```diff
@@ -132,6 +122,22 @@ pub(crate) fn var_set(
...
+        + handle.length as usize;
+
+    for (k, v) in data.vars.iter() {
+        size += k.len();
+        size += v.len();
+        size += std::mem::size_of::<String>() + std::mem::size_of::<Vec<u8>>();
+    }
+
+    // If the store is larger than the configured size, or 1mb by default, then stop adding things
+    if size > data.manifest.memory.max_var_bytes.unwrap_or(1024 * 1024) as usize && voffset != 0 {
```
nilslice, line 137: would it make sense to add a check for `data.manifest.memory.max_var_bytes.unwrap_or(1024 * 1024) as usize == 0` at the top of this function, with an error of something like `Vars are disabled by this host`, just to short-circuit this and leave a more clear error? 

i _think_ that the only error someone will see here is `Variable store is full`, which doesn't tell me if Vars are strictly disabled, which would mean I need to change my code as a plugin author. on the other hand, it also leaks some host-specific information to the plugin, which is probably ok?

zshipko: good idea, i will add that so the user can get a more helpful error message!

PR #696, opened 2024-03-18
feat: add Pool type for pooling plugin instances

zshipko:
- `Pool` allows for a configurable number of instances for each plugin and will block waiting for a plugin to become available if the limit is reached

### DIFF
--- a/runtime/src/pool.rs
+++ b/runtime/src/pool.rs
@@ -61,10 +61,7 @@ impl<Key: std::fmt::Debug + Clone + std::hash::Hash + Eq> Pool<Key> {
     }
 
     /// Add a plugin using a callback function
-    pub fn add<F: Fn() -> Result<Plugin, Error>>(&self, key: Key, source: F)
-    where
-        F: 'static,
-    {
+    pub fn add<F: 'static + Fn() -> Result<Plugin, Error>>(&self, key: Key, source: F) {
         let mut pool = self.inner.lock().unwrap();
         if !pool.instances.contains_key(&key) {
             pool.instances.insert(key.clone(), vec![]);
--- a/runtime/src/pool.rs
+++ b/runtime/src/pool.rs
@@ -1,38 +1,43 @@
+use crate::{Error, FromBytesOwned, Plugin, PluginBuilder, ToBytes};
 use dashmap::DashMap;
-use extism::{FromBytesOwned, ToBytes};
 
-#[derive(Clone)]
-pub struct PoolPlugin {
-    inner: std::rc::Rc<std::cell::RefCell<extism::Plugin>>,
-}
+/// `PoolPlugin` is used by the pool to track the number of live instances of a particular plugin
+#[derive(Clone, Debug)]
+pub struct PoolPlugin(std::rc::Rc<std::cell::RefCell<Plugin>>);
 
 impl PoolPlugin {
-    fn new(plugin: extism::Plugin) -> Self {
-        Self {
-            inner: std::rc::Rc::new(std::cell::RefCell::new(plugin)),
-        }
+    fn new(plugin: Plugin) -> Self {
+        Self(std::rc::Rc::new(std::cell::RefCell::new(plugin)))
     }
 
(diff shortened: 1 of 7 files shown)


runtime/src/pool.rs
```diff
@@ -0,0 +1,158 @@
...
+type PluginSource = dyn Fn() -> Result<Plugin, Error>;
+
+struct PoolInner<Key: std::fmt::Debug + Clone + std::hash::Hash + Eq = String> {
+    plugins: HashMap<Key, Box<PluginSource>>,
+    instances: HashMap<Key, Vec<PoolPlugin>>,
+}
+
+/// `Pool` manages threadsafe access to a limited number of instances of multiple plugins
+#[derive(Clone)]
+pub struct Pool<Key: std::fmt::Debug + Clone + std::hash::Hash + Eq = String> {
```
Nutomic, line 43: Im not sure what the `Key` parameter is good for. It may be useful to store multiple different plugin types in the same pool, but this is tricky as there is currently no way to get a list of all existing keys. Anyway the pool itself has very little overhead, so it makes more sense for users to write `HashMap<Key, Pool>` if needed. In my case Im using `Vec<Pool<()>>` because there is no need for any hashmap.

runtime/src/pool.rs
```diff
@@ -0,0 +1,158 @@
...
+    max_instances: usize,
+    inner: std::sync::Arc<std::sync::Mutex<PoolInner<Key>>>,
+}
+
+unsafe impl<T: std::fmt::Debug + Clone + std::hash::Hash + Eq> Send for Pool<T> {}
+unsafe impl<T: std::fmt::Debug + Clone + std::hash::Hash + Eq> Sync for Pool<T> {}
+
+impl<Key: std::fmt::Debug + Clone + std::hash::Hash + Eq> Pool<Key> {
+    /// Create a new pool
+    pub fn new(max_instances: usize) -> Self {
```
Nutomic, line 53: In most cases this should be fine with `max_instances = std::thread::available_parallelism()`. So I would consider using that value by default, removing the parameter here and providing a builder to set a different max value.

runtime/src/pool.rs
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+
+        loop {
+            if let Ok(Some(x)) = self.find_available(key) {
+                return Ok(Some(x));
+            }
+            if std::time::Instant::now() - start > timeout {
+                return Ok(None);
+            }
```
Nutomic, line 140: The loop should sleep for 100ms or so in each iteration.

zshipko:
@Nutomic thanks for the review! I like the suggestion about moving the map out of `Pool` - I will take a look at making some of these changes when I get a chance.

runtime/src/pool.rs
```diff
@@ -0,0 +1,158 @@
+use std::collections::HashMap;
+
+use crate::{Error, FromBytesOwned, Plugin, PluginBuilder, ToBytes};
+
+/// `PoolPlugin` is used by the pool to track the number of live instances of a particular plugin
+#[derive(Clone, Debug)]
+pub struct PoolPlugin(std::rc::Rc<std::cell::RefCell<Plugin>>);
```
Nutomic, line 7: This type should probably be exported with `pub use`, at the moment its inaccessible.

Nutomic:
Are you planning to finish this PR and merge it? Our corresponding feature is already merged using this code as git dependency (https://github.com/LemmyNet/lemmy/pull/5498). Our stable release is still some months away, by then it would be good to depend on a stable extism version with this included.

zshipko:
Hi @Nutomic - I think we could release this code, especially if it's already working for you. I will try to address your feedback and get this into a mergeable state in the next few weeks.

zshipko:
Hi @Nutomic - I've addressed most of your feedback, with the exception of removing the `Key` type parameter. I don't think I will be able to refactor to make that change. If you really need it I would be happy to review a PR based off of this branch, otherwise I think this should be ready!

PR #697, opened 2024-03-20
refactor!: update to wasmtime 19 

zshipko:
- Similar to https://github.com/extism/extism/pull/671 without wasip2 support
- Breaking: No longer copies Extism config values into WASI environment variables, these should be accessed using the Extism config functions instead

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -9,8 +9,8 @@ repository.workspace = true
 version.workspace = true
 
 [dependencies]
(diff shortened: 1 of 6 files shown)


nilslice:
Thanks!

On TCP/UDP access, let's just come to alignment about if this is good or not - even behind allowed_hosts - and understand the risks and potential ways to accidentally enable something undesirable.

zshipko:
Closing in favor of https://github.com/extism/extism/pull/723 to allow for more time to develop the socket/WASI command support.

PR #705, opened 2024-04-14
feat: add id() to CurrentPlugin

### LINKED ISSUE #704: Could we get access to the `id` on `CurrentPlugin`
I'd suggest just adding a readonly `id()` method.

ModProg:
fixes  #704


zshipko (approved):
Thanks!

PR #711, opened 2024-05-09
feat: per call context

chrisdickinson:
Add `plugin.call_with_context` and `current_plugin.context` methods, enabling per-call context to be looped from the guest invocation to any host functions it calls. In an HTTP server environment, this enables re-using a plugin across multiple requests while switching out backing connections and user information in host functions. (Imagine a host function, `update_user` -- previously the plugin would have to have been aware of the user to pass to the host function. Now that information is ambient.)

This is a backwards-compatible change and requires no changes to existing plugins.

Implement by adding a global, mutable externref to the extism kernel. Since most programming languages, including Rust, don't let you define these natively, we accomplish this by using `wasm-merge` to combine the kernel Wasm with Wasm generated by a WAT file containing only the global.

(This pattern might be useful for other Wasm constructs we can't use directly from Rust, like `v128` in argument parameters.)

Wasmtime requires extern refs to be `Any + Send + Sync + 'static`; we additionally add `Clone`. I haven't tried this with an `Arc` directly, but it should work at least for container structs that hold `Arc`'s themselves.


kernel/extism-context.wat
```diff
@@ -0,0 +1,3 @@
+(module
+  (global (export "extism_context") (mut externref) (ref.null extern))
```
chrisdickinson, line 2: We could add a layer of protection around this by saying that "we don't accept any Wasm module that imports `extism_context`." (This would protect against a malicious plugin that imports this externref, holds onto it across two calls, replacing the global value of the second call with the value from the first call.)

bhelx: Yeah my first thought is we don't want people to be able to poke this and use some other plugin-runner's context

runtime/src/tests/runtime.rs
```diff
@@ -307,6 +307,43 @@ fn test_toml_manifest() {
...
+        },
+    );
+
+    let mut plugin = Plugin::new(WASM_REFLECT, [f], true).unwrap();
+
+    let message = "hello world";
+    let output: String = plugin
+        .call_with_context(
+            "reflect",
+            "anything, really",
```
zshipko, line 337: hah, this was my initial thought - "then what's the input?" but I realized that the `context` value is only available in host functions (assuming we forbid access to the actual global somehow), which feels like a pretty powerful combination. this seems like a nice stepping-stone to allowing externref arguments while compilers are still catching up!

there could be some confusion around what to pass as input vs context (especially since we used to have a `Context` type that meant something totally different)

kernel/extism-context.wat
zshipko: returning an error if a plugin is trying to import the global directly sounds like a good idea - I also noticed that the context global gets reset before each call regardless of whether `call_with_context` is used, which makes me confident that a plugin couldn't access the context from another call.

runtime/src/plugin.rs
```diff
@@ -873,11 +893,35 @@ impl Plugin {
...
+        let data = input.to_bytes()?;
+        self.raw_call(&mut lock, name, data, Some(ExternRef::new(context)))
             .map_err(|e| e.0)
             .and_then(move |_| self.output())
     }
 
+    pub fn invoke(&mut self, name: &str, args: &[wasmtime::Val]) {
+        let lock = self.instance.clone();
+        let mut lock = lock.lock().unwrap();
+    }
```
zshipko, line 923: This seems unused?

runtime/src/tests/runtime.rs
zshipko: Maybe something like `HostContext` or `HostParam` might be more clear - open to other ideas too!

runtime/src/plugin.rs
chrisdickinson: Ooh, good catch, thank you!

runtime/src/tests/runtime.rs
chrisdickinson: Good point – I renamed to `host_context` / `HostContext` in the applicable places.

PR #715, opened 2024-05-17
fix(libextism): namespace ExtismValType

G4Vi:
Some of the old constants conflicted with Perl headers used for build Perl extensions such the [perl-sdk](https://github.com/extism/perl-sdk/blob/002e4437ac214c927f429f7088830f79d205e949/Extism/lib/Extism/XS.xs). This fix would allow removing the copy of extism.h (inlined in that file linked) out of the perl-sdk.

These constants especially `I32` likely have conflicts with many other large C codebases.

A new extism release with this fix should not be made until the effective sdks (cpp-sdk) have an update ready.

### DIFF
--- a/runtime/build.rs
+++ b/runtime/build.rs
@@ -7,7 +7,7 @@ fn main() {
 #define EXTISM_SUCCESS 0
 
 /** An alias for I64 to signify an Extism pointer */
-#define PTR I64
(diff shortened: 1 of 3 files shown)


zshipko (approved):
Looks good! This may affect some of the other SDKs, I can go through them next week

PR #718, opened 2024-05-22
refactor: remove HttpRequest.header alias

mhmd-azeez:
https://github.com/extism/dotnet-pdk/pull/84 brought to my attention that the Rust SDK supports both `headers` and `header` while the go sdk only supports `headers`. After talking to Zach, we decided to remove the `header` alias from the Rust SDK too, since it's obsolete and we want people to use `headers`


zshipko (dismissed):

zshipko:
There's also `memory.max` as an alias for `memory.max_pages` maybe it makes sense to get rid of that now too?

mhmd-azeez:
@zshipko done!

zshipko (approved):
Thanks!

PR #726, opened 2024-06-11
fix(libextism): improve static linking pkgconfig

G4Vi:
This includes two fixes:

* link static library by absolute path (should fix static linking on Mac/ with `lld`)
* link `libpthread` , this should be a no-op on systems with modern libc, but should fix static linking on systems with glibc pre 2.34 see https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread and error http://www.cpantesters.org/cpan/report/669fdfc8-25a4-11ef-9502-51fb6d8775ea

@zshipko if you could verify `make pkg-config-static` and the resulting `example` executable in `libextism` works on Mac that would be greatly appreciated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,8 +31,8 @@ endif
 build:
 	cargo build --release $(FEATURE_FLAGS) --manifest-path libextism/Cargo.toml $(TARGET_FLAGS)
 	sed -e "s%@CMAKE_INSTALL_PREFIX@%$(DEST)%" libextism/extism.pc.in > libextism/extism.pc
-	sed -e "s%@CMAKE_INSTALL_PREFIX@%$(DEST)%" libextism/extism-static.pc.in > libextism/extism-static.pc
-	sed -ie "s%@Libs.private: @%Libs.private: $(PRIVATE_LIBS)% " libextism/extism-static.pc
+	sed -e "s%@CMAKE_INSTALL_PREFIX@%$(DEST)%" \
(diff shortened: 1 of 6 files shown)


zshipko:
I just pushed a commit to fix the libextism example on macOS - the only thing I didn't fix is on macOS `-framework Security` also needs to be linked in the static pkg-config file - what do you think the best way of handling that is?

G4Vi:
Thanks! The best way of handling it is probably adding another variable to the template, `extism-static-pc.in`, and having the cli fill it in like it does for the prefix.

zshipko:
Alright I can add that - I think I can probably do something similar but avoid adding a variable so we maintain compatibility with existing CLI installations

zshipko:
Alright, this seems to be working now - let me know if you see anything that looks off @G4Vi 

I also have https://github.com/extism/cli/pull/94 ready to merge

G4Vi:
Looks good to me. The only thing I wonder about is `-framework Security` being part of `Libs` instead of `Libs.private`, but if it needs to go before libextism when it has to. Have you seen any other pkgconfig that uses it?

If you review I can merge without bypassing branch protections.

zshipko:
Including it in `Libs.private` still resulted in the linker error but `Libs` worked - I'm not sure why though. It could be the order of the libraries, but based on this search https://github.com/search?q=%22-framework+Security%22+path%3A*.pc&type=code it looks like people are using both `Libs` and `Libs.private`.

zshipko (approved):

PR #745, opened 2024-07-23
fix: require error messages to be null terminated in C SDK

zshipko:
Related to https://github.com/extism/python-sdk/issues/23 - there is currently no way to get the length of the error message, so we need to make sure it is a valid C string.


nilslice:
great catch - thanks for the quick fix!

PR #758, opened 2024-08-27
feat(runtime): add log level set/get for pdk control

nilslice:
In addition to this change, @zshipko is looking into if we need to extract basic log level info from the more complex `filter` string, and how to require `tracing` compatibility to at least set some level (or to work when there is no level set thus ignoring all logs)

For other runtimes, there is still an equivalent implementation for `get_log_level() -> I32`  in Go and JS, as well as the ability to set the log level for the host, which is then loaded when `get_log_level` is called. 

TODO:
- [ ] Go SDK implementation
- [ ] JS SDK implementation

### DIFF
--- a/runtime/extism.h
+++ b/runtime/extism.h
@@ -313,5 +313,5 @@ bool extism_plugin_reset(ExtismPlugin *plugin);
 const char *extism_version(void);
 
 #ifdef __cplusplus
-} // extern "C"
-#endif // __cplusplus
+}  // extern "C"
+#endif  // __cplusplus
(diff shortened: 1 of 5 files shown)


nilslice:
@zshipko thank you for the follow up commit. I am doing a quick spike on the Go SDK implementation to see if there is any complexity as far as consistency to this impl goes. 

Will update here and we can merge this after that.

nilslice:
Ok, there's a tiny bit of complexity 😆, but no more than to be expected... LGTM!

zshipko (approved):

PR #760, opened 2024-08-30
fix: Vec.as_ptr() might return a dangling pointer

evacchi:
Both [as_ptr][as_ptr] and [as_mut_ptr][as_mut_ptr] are allowed to return a dangling raw pointer when the Vec size is 0.

The idea is that you should guard that read checking the size. This probably works well in most cases, but at the very least in the java-sdk, the JNA machinery tries to be helpful and it dereferences the pointer, causing a SIGSEGV.

The solution is to check if the resulting vector is empty and return null instead. A new, empty vector would be better, but I think that would not solve the problem, because the problem is caused by a new, empty vector in the first place.

Caveat: this might break consumers downstream.
On the other hand: consumers that do not check for the nInput, nOutput counts are just waiting to explode, like JNA.

This addresses https://github.com/extism/java-sdk/issues/27

[as_ptr]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_ptr
[as_mut_ptr]: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_mut_ptr

### DIFF
--- a/runtime/src/sdk.rs
+++ b/runtime/src/sdk.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::missing_safety_doc)]
 
-use std::os::raw::c_char;
+use std::{os::raw::c_char, ptr::null_mut};
 
 use crate::*;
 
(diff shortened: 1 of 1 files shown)


evacchi:
I am not sure what's wrong with CI 😅  WorksOnMyMachine™️

zshipko (approved):

PR #761, opened 2024-08-30
fix: patch regalloc breaking change

chrisdickinson:
regalloc2 released a breaking change in a patch version [1].

thanks to @adunne09 for identifying this and providing a fix!

[1]: https://github.com/bytecodealliance/regalloc2/commit/98fbea5bac712a955ed478d5527f6896b65a7d18


nilslice (dismissed):
I think there are a few other things in the pipeline that we could release along with this to quickly roll out the fix

Namely: 
- #760
- #759
- #758

PR #762, opened 2024-09-03
fix: better error handling when plugin runs out of fuel

zshipko:
- Checks error results to determine if a plugin ran out of fuel, if the fuel is 0 after we get the error, then we return an out of fuel error message instead.
- Updates `extism_plugin_error` to check `Plugin::error_msg` regardless of the kernel's error state

### DIFF
--- a/runtime/src/current_plugin.rs
+++ b/runtime/src/current_plugin.rs
@@ -1,3 +1,5 @@
+use anyhow::Context;
+
 use crate::*;
 
 /// CurrentPlugin stores data that is available to the caller in PDK functions, this should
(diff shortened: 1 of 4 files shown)


zshipko:
@evacchi - this should fix the segfault you were running into. it seems like the fuel was running out before the plugin could set an error which was causing it to get into a weird state.

evacchi (approved):
This change effectively causes my plugin to consistently fail with `plugin ran out of fuel` for all values below a certain threshold, including down to 0 👍 let's also merge #760 so that we can depend on it in the java-sdk

PR #763, opened 2024-09-17
feat: add ability to configure static memory size

zshipko:
This makes it possible to configure the amount of memory that's allocated up front when instantiating a plugin.

### DIFF
--- a/runtime/src/plugin_builder.rs
+++ b/runtime/src/plugin_builder.rs
@@ -40,6 +40,7 @@ pub struct PluginBuilder<'a> {
     debug_options: DebugOptions,
     cache_config: Option<Option<PathBuf>>,
     fuel: Option<u64>,
+    static_memory_size: Option<u64>,
 }
 
 impl<'a> PluginBuilder<'a> {
@@ -52,6 +53,7 @@ impl<'a> PluginBuilder<'a> {
             debug_options: DebugOptions::default(),
             cache_config: None,
             fuel: None,
+            static_memory_size: None,
         }
     }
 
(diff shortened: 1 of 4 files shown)


nilslice:
Nice! What happens if this limit is exceeded? And is it only upon instantiation?

zshipko:
If the limit is exceeded then it switches to dynamic memory, where more memory is allocated and the old memory is copied over - so there is some performance downside to growing memory at that point. More information here: https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.static_memory_maximum_size

runtime/src/plugin_builder.rs
```diff
@@ -150,12 +152,19 @@ impl<'a> PluginBuilder<'a> {
...
-    // Limit the number of instructions that can be executed
+    /// Limit the number of instructions that can be executed
     pub fn with_fuel_limit(mut self, fuel: u64) -> Self {
         self.fuel = Some(fuel);
         self
     }
 
+    /// Limit the size of the memory allocated up-front when
+    /// instantiating a module
+    pub fn with_static_memory_size(mut self, size: u64) -> Self {
```
SebastianHambura, line 163: Considering the potential performance implications, and the big warnings from the underlaying wasmtime function's documentation, maybe it would be worth it to add a link to https://docs.wasmtime.dev/api/wasmtime/struct.Config.html#method.static_memory_maximum_size in this function's documentation

zshipko:
I just opened https://github.com/extism/extism/pull/764, which solves the same problem in a more general way

PR #765, opened 2024-09-18
cleanup(kernel): only try to re-use free blocks before memory.grow

zshipko:
Updates the kernel to only scan for free blocks before a `memory.grow` operation, this should improve performance of `alloc` by not iterating over every block for each allocation.


zshipko:
Just realized this would also change the whole kernel test suite - I will re-open this after doing some more benchmarking to see if it's worth it.

zshipko:
Comparing an example that performs 10,000 allocation with the current implementation on main shows that this branch is a significant improvement:

```
allocations/allocations time:   [153.68 ms 157.70 ms 161.94 ms]
                        change: [-75.472% -74.562% -73.548%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #776, opened 2024-10-09
fix(plugin_call): set rc to EXIT_SIGNALED_SIGABRT when wasmtime bails out on plugin call

### LINKED ISSUE #775: libextism `extism_plugin_call` sometimes returns 0 on error.
This happens when [wasmtime trap](https://docs.rs/wasmtime/latest/wasmtime/enum.Trap.html) occurs during execution, for example when an `unreachable` statement is executed.

At least two of the libextism using sdks expect success when a zero is returned: [cpp-sdk](https://github.com/extism/cpp-sdk/blob/fd09b0db7f30a6c2c8d374048a44d4faac542a83/src/plugin.cpp#L79-L86) and [perl-sdk](https://github.com/extism/perl-sdk/blob/4c4ce5171b95df0918bebf9a4f734926528d6b54/Extism/lib/Extism/Plugin.pm#L70-L73)

The [python-sdk](https://github.com/extism/python-sdk/blob/16273fb12f2e5c2df33702398f221227e348e7e1/extism/extism.py#L510-L514) uses the presence of an error string also to determine if there's an error.

I can update the sdks to check for an error string instead, but I think it would be preferable to avoid the need to call `extism_plugin_error` every time you call a plugin function. The existing documentation for [extism_plugin_call](https://github.com/extism/extism/blob/2bf391f2362cb06aa984d02992edb1f9af497ff6/libextism/API.md#extism_plugin_call) states `Returns 0 when the call is successful.`

G4Vi:
Fixes #775

134/EXIT_SIGNALED_SIGABRT was chosen as the wasmtime CLI exits with it executes `unreachable`.

If it would be preferable, I could make the modification just to `extism_plugin_call_with_host_context` instead.


runtime/src/plugin.rs
```diff
@@ -904,6 +904,10 @@ impl Plugin {
                         )));
                     }
                 }
+            } else if let Err(e) = &res {
+                if e.is::<wasmtime::Trap>() {
+                    rc = 134; // EXIT_SIGNALED_SIGABRT
+                }
```
zshipko, line 910: I wonder if we should do something like:

```suggestion
                } else if rc == 0 {
                  rc = -1;
                }
```

in case it's possible for there to be another case where an error is returned but rc == 0. But maybe that's not needed.

zshipko (approved):
Looks good! One comment, but feel free to ignore it.

runtime/src/plugin.rs
G4Vi: In bf95d06 I reworked it to default to -1 and avoid hiding errors that can happen first. Extism errors can happen before wasmtime errors, which happen before retrieving output errors.

zshipko (approved):

PR #783, opened 2024-10-28
chore: update to wasmtime 26

zshipko:


bhelx (approved):
Glad to see it's such a simple upgrade. Curious if we know the performance implications of this.

runtime/Cargo.toml
```diff
@@ -9,9 +9,9 @@ repository.workspace = true
 version.workspace = true
 
 [dependencies]
-wasmtime = ">= 20.0.0, < 24.0.0"
-wasi-common = ">= 20.0.0, < 24.0.0"
-wiggle = ">= 20.0.0, < 24.0.0"
+wasmtime = {version = ">= 26.0.0, < 27.0.0"}
```
bhelx, line 12: this is a pretty tight range. curious if there are any implications to only supporting this one version.

zshipko: yeah it's tricky - I would be open to letting this sit until the next wasmtime release so we can support at least 2 versions but also don't want to fall too far behind. But I'm definitely not in a rush to merge this right now.

I'm not sure it's an issue for most applications to only have one supported version since I think we are the ones pulling in the wasmtime dependency is most cases. But there could be some projects out there that depend on a specific wasmtime version too.

zshipko:
Looks like there are some minor vulnerabilities in wasmtime v23 according to https://github.com/extism/extism/security/dependabot/15 - I think it's worth merging this now and hopefully we can expand the bounds from there as new wasmtime releases are made.

PR #794, opened 2024-11-25
fix: remove unwrap() from extism_compiled_plugin_new

chrisdickinson:
`extism_compiled_plugin_new` would panic on bad manifest input (there's a test in the python sdk [1] that hits this directly.) Catch the error and set `errmsg` appropriately instead.

Additionally: golf the function pointer casting for the various `plugin_new` calls, with the goal of reducing the number of allocations by leaning on iterator size hints (and reducing the scope of `unsafe{}` use.) This introduces a breaking change. Previously `NULL` values were silently accepted in the `functions**` list; now they fail. This maintains consistency with behavior on "consumed" `ExtismFunction` pointer.

[1]: https://github.com/extism/python-sdk/blob/main/tests/test_extism.py#L50-L53


zshipko (approved):
Looks great, thanks!

PR #796, opened 2024-12-01
Change `function_exists` to `&self`

milesj:
I wrap the `Plugin` instance in a `RwLock` and because `function_exists` requires `&mut self`, I have to acquire a write lock everytime to just check if a function exists, which causes lock contention.

This changes it to `&self` since it doesn't need to mutate.


bhelx:
Thank you :)

Will have @zshipko peek at it on Monday

zshipko (approved):
Thanks! Test failures are clippy related, I will fix those in a follow up PR.

Issue #801, opened 2024-12-02
improve error messages around failed imports


zshipko:
During the linking step we should attempt to detect missing imports or type mismatches in order to provide better error messages during instantiation.

PR #806, opened 2024-12-13
fix: improve sdk error messages around imports

zshipko:
Closes https://github.com/extism/extism/issues/801

Updates the error from:

```
Unable to compile Extism plugin: incompatible import type for `extism:host/user::wrong_type`
```

to:

```
Unable to compile Extism plugin: types incompatible: expected type `(func (result i32))`, found type `(func (result i64))`
```

### DIFF
--- a/runtime/src/plugin.rs
+++ b/runtime/src/plugin.rs
@@ -298,33 +298,6 @@ fn relink(
     let mut linker = Linker::new(engine);
     linker.allow_shadowing(true);
 
-    for (name, module) in modules.iter() {
-        if name == EXTISM_ENV_MODULE {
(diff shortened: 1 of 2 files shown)


nilslice (dismissed):
nice!

zshipko:
Just pushed another change to update the error message when an unknown import from `extism:host/env` is found:

```
Unable to compile Extism plugin: Invalid function import from extism:host/env: something_that_doesnt_exist (type (func (result i32)))

Note: This may indicate that the PDK that was used to build this plugin has additional features that aren't available in this version of the SDK, try updating the SDK to the latest version.
```

PR #807, opened 2024-12-16
feat: added a function to track fuel consumption

pwnintended:
This PR adds a function to get the fuel consumed by a plugin. This is useful for systems that want to fairly balance compute. It's my first time doing anything in rust so let me know if something needs improvements!

### DIFF
--- a/runtime/src/plugin.rs
+++ b/runtime/src/plugin.rs
@@ -147,7 +147,6 @@ pub struct Plugin {
     pub(crate) error_msg: Option<Vec<u8>>,
 
     pub(crate) fuel: Option<u64>,
-    pub(crate) remaining_fuel: Option<u64>,
 
     pub(crate) host_context: Rooted<ExternRef>,
 }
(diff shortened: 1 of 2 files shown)


zshipko:
Thanks for working on this! I think this could be very helpful, one thing worth noting is the fuel is currently reset for before each call - I'm not sure if this will be an issue for your use case.

I have a branch with a slightly simplified implementation here: https://github.com/extism/extism/tree/consumed-fuel-direct - it avoids the extra field on `Plugin` by just calling `store.get_fuel` directly in the `fuel_consumed` - is there a benefit to storing the remaining fuel in the `Plugin` structure that I'm missing?

pwnintended:
We actually want fuel to be reset for each call, the usecase for us is to see how much fuel is consumed by a call so that we can keep track of this.

I don't see a reason to store it if it's reset after every call. Calling it directly like in your branch is better I think.

zshipko:
Great, I will push up my changes to your branch and merge once CI passes!

PR #824, opened 2025-01-29
Uppercase the http method

bhelx:


runtime/src/pdk.rs
```diff
@@ -233,7 +233,10 @@ pub(crate) fn http_request(
             )));
         }
 
-        let mut r = ureq::request(req.method.as_deref().unwrap_or("GET"), &req.url);
+        // default to GET and uppercase
+        let http_method = req.method.unwrap_or("GET".into()).to_uppercase();
+
```
bhelx, line 238: as discuss in chat. some apis and http servers are sensitive to this for some reason so we're going to force it for the client.

zshipko (approved):

PR #836, opened 2025-02-26
fix: throw error on reentrant plugin call

chrisdickinson:
Previously, code that called a plugin that called a host function that called back into the plugin would hang in an attempt to attain the lock. Switch to `try_lock` so we can surface the error.

(Just noticed this in Discord this AM.)


zshipko (approved):
Thanks, this will be a nice improvement!

Issue #714, opened 2024-05-15
getting a list of a plugin’s exported functions


Somfic:
Hi.

I was wondering if it’d be possible to, given a `Plugin`, figure out which functions are exported by it?

nilslice:
Hey - we don't officially support this via Extism, but you can use another tool we make, Modsurfer. Either statically, via the web[0] or CLI[1], or dynamically in a program using it as an Extism plugin[2].

[0]: https://modsurfer.dylibso.com
[1]: https://github.com/dylibso/modsurfer
[2]: https://modsurfer.dylibso.com/module?hash=0c20c61f67108ebccae1db0be6df7c7d14b2567d5606154278ee390f43e1f408

For [2], via an Extism plugin, you would call the export `parse_module`, passing it the bytes of your wasm plugin, and in return, the plugin will provide a `Module` protobuf. 

We provide Go and Rust libraries, but you can also generate your own protobuf library in any language they support using the `api.proto` as the definition. See the files here: https://github.com/dylibso/modsurfer/tree/main/proto/v1

Somfic:
Awesome thanks for the quick help!

nilslice:
For sure! If you have any issues with these approaches, please let us know. I might also add another export to this plugin (would be a new .wasm) which would just list the exports by name in a JSON list `["count_vowels", "etc"]` so you don't need to deal with the protobuf types... would that help?

nilslice:
@Somfic -- ok, the Extism plugin[0] (new version to download) has an export called `list_extism_funcs`, which just returns the exports that you can call from Extism specifically.

[0]: https://modsurfer.dylibso.com/module?hash=67a67abb2a863ba45967836b9ba75ce9b287c7e5e1a86ff02cfcb27dd41350f4

The usage is like: 

```sh
cat plugin.wasm | extism call modsurfer.wasm list_extism_funcs --stdin | jq 
["export_1", "export_2", ...]
```

Somfic:
Will use the `modsurfer` plugin for now then! :)

PR #837, opened 2025-02-28
Add: Added a function to list functions in a Plugin.

RiadYan:
This is just a quick function that I added because I had to use it in my code.
I am not sure why it has not been added yet but if you need more details I will be happy to provide them.
And sorry for spamming the fork button, I was narrow minded and brutally tried to reproduce what has been done in the (go-sdk) before realizing that it isn't useful.

If needed I can add a struct that contains the name the params and result.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -40,7 +40,8 @@ pub use extism_convert::{FromBytes, FromBytesOwned, ToBytes};
 pub use extism_manifest::{Manifest, Wasm, WasmMetadata};
 pub use function::{Function, UserData, Val, ValType, PTR};
 pub use plugin::{
-    CancelHandle, CompiledPlugin, Plugin, WasmInput, EXTISM_ENV_MODULE, EXTISM_USER_MODULE,
+    CancelHandle, CompiledPlugin, FunctionDefinition, Plugin, WasmInput, EXTISM_ENV_MODULE,
+    EXTISM_USER_MODULE,
 };
 pub use plugin_builder::{DebugOptions, PluginBuilder};
 
--- a/runtime/src/plugin.rs
+++ b/runtime/src/plugin.rs
@@ -579,6 +579,27 @@ impl Plugin {
             .unwrap_or(false)
     }
 
+    /// List all functions names inside of the plugin.
+    pub fn list_functions(&self) -> Vec<String> {
+        self.modules[MAIN_KEY]
+            .exports()
+            .filter_map(|export| {
+                if let wasmtime::ExternType::Func(f) = export.ty() {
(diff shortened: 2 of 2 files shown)


nilslice:
@RiadYan - Thanks, but we would prefer that this is something you do in your own application and have discussed it quite a lot in other issues.

See how you can leverage it: https://github.com/extism/extism/issues/714

> I am not sure why it has not been added yet 

Because this is not the only implementation of Extism we maintain, we would need to add it to every single API to remain consistent. It's not worth the maintenance effort, when it can be easily done via an external plugin.

RiadYan:
> Thanks for the PR! Being able to list functions is definitely something that comes up every once in a while - I'm curious what your use case is?

For mass verification of functions, I want to limit the exported functions and check every one of them and their returns to make sure nothing goes unnoticed. The function_exists is here but if I want to check the return and it doesn't allow me to do so sadly.
Also, if I want to check 50 functions it starts to get heavy so I just need a list. I also wanted to allow the users to see which functions are given by the plugin they just installed, as a pledge of transparency.

> Because this is not the only implementation of Extism we maintain, we would need to add it to every single API to remain consistent. It's not worth the maintenance effort, when it can be easily done via an external plugin.

Thank you for the insight, I read the Issue #714 before starting to write the PR but saw also that it has somewhat been added to the [go-sdk](https://github.com/extism/go-sdk/pull/79/files#diff-c95a9381cb3d755657cf984eba630f599695529c52a3b407d902d38bcea81f8a), I thus thought it was just forgotten. 
Thank you for the quick reply though, I appreciate your work and I am happy to see the community really alive !
Much love and good luck <3

zshipko:
It sounds like modsurfer would be a great fit for your use-case, particularly the dynamic usage (#2) mentioned in this message https://github.com/extism/extism/issues/714#issuecomment-2111475055

Using modsurfer would also allow you to verify the module before you actually load it into your application

RiadYan:
> It sounds like modsurfer would be a great fit for your use-case, particularly the dynamic usage (#2) mentioned in this message [#714 (comment)](https://github.com/extism/extism/issues/714#issuecomment-2111475055)
> 
> Using modsurfer would also allow you to verify the module before you actually load it into your application

I have taken a look at it previously, I want to keep my app lean and cross-platform and offline, modsurfer looks great though, if it fits my needs I will possibly consider it, but I will try it out nonetheless.
Thank you for your reply again !
Good luck and I wish you all the best <3

nilslice:
@RiadYan - just to clarify, @zshipko meant that .wasm module (which is the same one used in the Modsurfer web app). 

You can use it to list the extism functions in other .wasm modules all offline, in-process. It's just another Extism plugin that you would call from your Host code.

PR #839, opened 2025-03-21
doc: explain how to see plug-in logs

Pascal-So:
I know that there is already a logging example in `runtime/examples/log_callback.rs` but I think that's not where people go to look to figure out how the logging functionality works. At least for me that wasn't the case :)


zshipko (approved):
Thanks!

PR #869, opened 2025-07-04
Add `Pool.function_exists` with caching

Nutomic:
This wrapper caches the result of `plugin.function_exists`, to avoid having to load a plugin from the pool every single time just to find out if the given function exists. It can improve performance if there are many plugin hooks without corresponding plugin functions.


zshipko:
Looks like there are some unrelated clippy errors, I will take a look at fixing those today.

zshipko:
Just pushed a commit for the clippy changes, I will merge once CI passes.

PR #876, opened 2025-07-22
[WIP] new: Support structured tracing logs.

milesj:
Extism has logging/tracing support currently through explicit macro usage, but what it doesn't support is inheriting `tracing` logs that come from 3rd-party crates. Right now all those messages are lost, which is a huge bummer.

I wanted to figure it if inheriting these 3rd-party logs is even possible... and it is! Turns out you can set `tracing::set_global_default` in WASM to read these logs, and then pipe them to extism. Here's an example I've been experimenting with: https://github.com/moonrepo/proto/blob/master/crates/warpgate-pdk/src/tracing.rs#L98

While this works, there is a major downside, and that is that field information is lost. Extism only supports the message string, so to pipe field data through, you need to include it in the message. This results in ugly logs, as seen here (notice the `extism::pdk` lines):

Screenshot 2025-07-16 at 3 00 27 PM
Screenshot 2025-07-16 at 2 42 41 PM

## Solution

So in this PR, I wanted to figure out a way to support this field data, and a custom target (to override `extism::pdk`). I'm doing this by introducing a new struct, `TracingLog`, that includes this information. I've stored this struct in `extism-convert` so that it can also be used in the PDK (there will be a sibling PR to this one).

I then use the `tracing-dynamic` crate (https://github.com/BrynCooke/tracing-dynamic) to send the event, because it's currently impossible to set dynamic fields on an event (why tracing why???).

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -28,6 +28,7 @@ serde_json = "1"
 toml = "0.9"
 sha2 = "0.10"
 tracing = "0.1"
+tracing_dynamic = "0.3"
 tracing-subscriber = { version = "0.3.18", features = [
   "std",
   "env-filter",
@@ -42,13 +43,16 @@ uuid = { version = "1", features = ["v4"] }
 libc = "0.2"
 
 [features]
-default = ["http", "register-http", "register-filesystem", "wasmtime-default-features"]
-register-http = ["ureq"]                                   # enables wasm to be downloaded using http
-register-filesystem = []                                   # enables wasm to be loaded from disk
-http = ["ureq"]                                            # enables extism_http_request
-wasmtime-default-features = [
-  'wasmtime/default',
+default = [
+  "http",
+  "register-http",
+  "register-filesystem",
+  "wasmtime-default-features",
 ]
+register-http = ["ureq"] # enables wasm to be downloaded using http
+register-filesystem = [] # enables wasm to be loaded from disk
+http = ["ureq"] # enables extism_http_request
+wasmtime-default-features = ['wasmtime/default']
 
 [build-dependencies]
 cbindgen = { version = "0.29", default-features = false }
--- a/convert/src/lib.rs
+++ b/convert/src/lib.rs
@@ -11,10 +11,10 @@ extern crate self as extism_convert;
 pub use anyhow::Error;
 
 mod encoding;
-
 mod from_bytes;
 mod memory_handle;
 mod to_bytes;
+mod tracing_log;
 
 pub use encoding::{Base64, Json};
 
@@ -33,6 +33,7 @@ pub use encoding::Raw;
 pub use from_bytes::{FromBytes, FromBytesOwned};
 pub use memory_handle::MemoryHandle;
 pub use to_bytes::ToBytes;
+pub use tracing_log::TracingLog;
 
 #[cfg(test)]
 mod tests;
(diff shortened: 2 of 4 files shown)


runtime/Cargo.toml
```diff
@@ -42,13 +43,16 @@ uuid = { version = "1", features = ["v4"] }
 libc = "0.2"
 
 [features]
-default = ["http", "register-http", "register-filesystem", "wasmtime-default-features"]
-register-http = ["ureq"]                                   # enables wasm to be downloaded using http
-register-filesystem = []                                   # enables wasm to be loaded from disk
-http = ["ureq"]                                            # enables extism_http_request
-wasmtime-default-features = [
-  'wasmtime/default',
+default = [
```
milesj, line 46: my editor did this

zshipko (commented):
this is neat! it looks like a great improvement. i will test it out a little more at some point today, but i don't see why not to move forward with this since it won't immediately require PDK changes beyond what you've done in the rust-pdk

milesj:
@zshipko I haven't tested this yet because I can't compile the project:

```
   Compiling extism-sys v0.0.0+replaced-by-ci (/Users/miles/Projects/extism/extism-maturin)
error: failed to run custom build command for `extism-sys v0.0.0+replaced-by-ci (/Users/miles/Projects/extism/extism-maturin)`

Caused by:
  process didn't exit successfully: `/Users/miles/.cargo/shared-target/debug/build/extism-sys-5924b66f727e8df4/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=src/extism.c
  cargo:rerun-if-changed=../runtime/extism.h
  OPT_LEVEL = Some(0)
  OUT_DIR = Some(/Users/miles/.cargo/shared-target/debug/build/extism-sys-c4948024f0f7da27/out)
  TARGET = Some(aarch64-apple-darwin)
  HOST = Some(aarch64-apple-darwin)
  cargo:rerun-if-env-changed=CC_aarch64-apple-darwin
  CC_aarch64-apple-darwin = None
  cargo:rerun-if-env-changed=CC_aarch64_apple_darwin
  CC_aarch64_apple_darwin = None
  cargo:rerun-if-env-changed=HOST_CC
  HOST_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
  RUSTC_WRAPPER = None
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some(true)
  cargo:rerun-if-env-changed=MACOSX_DEPLOYMENT_TARGET
  MACOSX_DEPLOYMENT_TARGET = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = None
  cargo:rerun-if-env-changed=HOST_CFLAGS
  HOST_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64_apple_darwin
  CFLAGS_aarch64_apple_darwin = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64-apple-darwin
  CFLAGS_aarch64-apple-darwin = None
  CARGO_ENCODED_RUSTFLAGS = Some()

  --- stderr

  thread 'main' panicked at extism-maturin/build.rs:56:48:
  called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
warning: build failed, waiting for other jobs to finish...
```

milesj:
This is just `cargo check --workspace` from the root.

PR #892, opened 2026-02-11
Fix pool thread safety issues

### LINKED ISSUE #891: Pool implementation is not thread safe
`PoolInner`'s `unsafe` implementation of `Sync` is unsound and can cause runtime borrow check panic.

https://github.com/extism/extism/blob/7230657f6f4d8ce5efee291f171ca45782d8c3fe/runtime/src/pool.rs#L79

The `PoolInner` struct has a vec of `PoolPlugins`:

https://github.com/extism/extism/blob/7230657f6f4d8ce5efee291f171ca45782d8c3fe/runtime/src/pool.rs#L42-L44

https://github.com/extism/extism/blob/7230657f6f4d8ce5efee291f171ca45782d8c3fe/runtime/src/pool.rs#L73-L76

As commented in the code, "`PoolPlugin` is used by the pool to track the number of live instances of a particular plugin". This tracking is done by checking the `Rc` refcount. This means that the code has to carefully lock around every read or modification of the `Rc` to soundly implement `Sync`.

One problem, however, is that `PoolPlugin` does not implement its own `Drop`. This means that when a `PoolPlugin` is dropped, the refcount decrement is not synchronized, which can result in a runtime borrow check panic. (In fact, `PoolPlugin` has no internal synchronization; it relies entirely on external locking. This means there's no trivial fix to synchronize just the `Drop` - the synchronization strategy needs to change fundamentally (e.g., replacing `Rc<RefCell<Plugin>>` with `Arc<Mutex<Plugin>>`)).

I found this issue when writing benchmarks for my project that uses Extism pools. The panic is intermittent, but it manifests like this:

```
thread 'tokio-runtime-worker' (51141255) panicked at crates/extism/runtime/src/pool.rs:53:16:
RefCell already borrowed

thread 'main' (51141232) panicked at crates/executor/benches/plugin_execution.rs:157:55:
Task panicked: JoinError::Panic(Id(339173), "RefCell already borrowed", ...)
```

I wasn't able to create a simple Extism unit test or a simple independent test that reproduces the issue. (I could put more effort into this if requested), but I implemented a fix using `Arc<Mutex<Plugin>>` instead of `Rc<RefCell<Plugin>>` in `pool.rs` and verified that I can no longer reproduce the panic, and that the change doesn't cause any measurable perf penalty in my own benchmarks. I can submit my fix as a PR if that sounds like an acceptable fix.

bgmerrell:
This is an alternative to #893 

The pool implementation had an unsound `unsafe impl Sync` on `PoolInner` containing `Vec<PoolPlugin>` where `PoolPlugin` was `Rc<RefCell<Plugin>>`. This caused intermittent "RefCell already borrowed" panics in multi-threaded usage (issue #891).

The problem with the original design was that it seemed to assumed external locking (the pool's `Mutex`) would be sufficient to make `Rc<RefCell<>>` safe to share across threads. However, there were several issues that made the design unsound:

1. `PoolPlugin` had no `Drop` implementation to synchronize refcount decrements. When dropped on different threads, unsynchronized `Rc` refcount modifications could lead to data races.
2. The pool used `Rc::strong_count == 1` to track availability, but had no control over when clones were created or destroyed. Callers could clone and keep `PoolPlugin` instances, breaking the availability tracking.
3. The `PoolPlugin` type didn't leverage the type system to enforce safe usage patterns; it relied entirely on external locking that the type system couldn't verify.

This replaces the unsafe implementation with proper thread-safe primitives and move semantics inspired by the deadpool crate.

Changes:

- Replace shared `Rc<RefCell<Plugin>>` with move semantics: plugins are moved out of a `VecDeque` when checked out and moved back on drop
- `PoolPlugin` now owns the plugin and implements `Drop` to automatically return it, making checkout/return atomic and type-safe.
- Add `Send` bound to `UserDataHandle::Rust` and add `T: Send` requirements to
  `UserData::new()` and `Function::new()`.  Previously, `UserData<T>` was marked
  `Send` via  `unsafe impl<T> Send` regardless of whether `T` was `Send`. When
  converting   `UserData<T>` to `UserDataHandle::Rust`, the type is erased to
  `Arc<Mutex<dyn Any>>`.  Adding `+ Send` to the trait object preserves the `Send`
  constraint through type erasure, but the compiler enforces that the cast is only valid
  when `T: Send`. The explicit bounds ensure this is checked at the API boundaries.
  This makes `PluginBuilder` with custom functions properly `Send`, enabling
  pools with custom host functions. **I don't know if this is OK, I think there is likely
  something in I am not fully grokking in the current design with these unsafe Send+Sync impls**
- Replace busy-wait polling with `Condvar` for efficient waiting
- `PoolPlugin` implements `Deref`/`DerefMut` for ergonomic access
- Use `Weak` reference to pool so checked-out plugins don't keep pool alive
- Remove all `unsafe impl Send/Sync`. Soundness is now enforced by the type system.
- Fix `function_exists` to handle timeout gracefully instead of panicking.
- Add test for pools with captured `PluginBuilder` containing custom functions.

Breaking API changes:

- `PoolPlugin` no longer has a `.plugin()` method - it now implements
  `Deref`/`DerefMut`, so you can use it directly where you previously
  called `.plugin()`. Example: `pool.get()?.call(...)` instead of
  `pool.get()?.plugin().call(...)`
- `PoolPlugin::call()` now takes `&mut self` instead of `&self` (direct
  ownership instead of interior mutability via `RefCell`)
- `UserData` types must be `Send` (previously claimed via `unsafe impl`
  but not enforced through type erasure to `dyn Any`)

Fixes #891

### DIFF
--- a/convert-macros/tests/ui/invalid-encoding.stderr
+++ b/convert-macros/tests/ui/invalid-encoding.stderr
@@ -1,6 +1,7 @@
 error: encoding needs to be specified
 
          = try: `#[encoding(Json)]`
+
  --> tests/ui/invalid-encoding.rs:3:10
   |
 3 | #[derive(ToBytes)]
@@ -12,6 +13,7 @@ error: expected attribute arguments in parentheses: #[encoding(...)]
 
          = note: expects a path
          = try: `#[encoding(Json)]`
+
  --> tests/ui/invalid-encoding.rs:7:3
   |
 7 | #[encoding]
@@ -21,6 +23,7 @@ error: expected parentheses: #[encoding(...)]
 
          = note: expects a path
          = try: `#[encoding(Json)]`
+
   --> tests/ui/invalid-encoding.rs:11:12
    |
 11 | #[encoding = "string"]
@@ -30,6 +33,7 @@ error: unexpected token
 
          = note: expects a path
          = try: `#[encoding(Json)]`
+
   --> tests/ui/invalid-encoding.rs:15:21
    |
 15 | #[encoding(something, else)]
@@ -38,6 +42,7 @@ error: unexpected token
 error: only one encoding can be specified
 
          = try: remove `#[encoding(Encodings)]`
+
   --> tests/ui/invalid-encoding.rs:20:1
    |
 20 | #[encoding(Encodings)]
--- a/runtime/src/function.rs
+++ b/runtime/src/function.rs
@@ -80,7 +80,7 @@ pub(crate) enum UserDataHandle {
     #[allow(dead_code)]
     C(Arc<CPtr>),
     #[allow(dead_code)]
-    Rust(Arc<std::sync::Mutex<dyn std::any::Any>>),
+    Rust(Arc<std::sync::Mutex<dyn std::any::Any + Send>>),
 }
 
 /// UserData is used to store additional data that gets passed into host function callbacks
@@ -94,7 +94,7 @@ pub enum UserData<T: Sized> {
     Rust(Arc<std::sync::Mutex<T>>),
 }
 
-impl<T: Default> Default for UserData<T> {
+impl<T: Default + Send> Default for UserData<T> {
     fn default() -> Self {
         UserData::new(T::default())
     }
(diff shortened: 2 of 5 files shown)


bgmerrell:
https://github.com/extism/extism/pull/893 was merged instead

PR #893, opened 2026-02-11
Fix pool thread safety issues (option 2)

### LINKED ISSUE #891: Pool implementation is not thread safe

bgmerrell:
This is an alternative to #892 

The pool implementation had an unsound `unsafe impl Sync` on `PoolInner` containing `Vec<PoolPlugin>` where `PoolPlugin` was `Rc<RefCell<Plugin>>`. This caused intermittent "RefCell already borrowed" panics in multi-threaded usage (issue #891).

The problem with the original design was that it assumed external locking (the pool's `Mutex`) would be sufficient to make `Rc<RefCell<>>` safe to share across threads. However, there were several issues that made the design unsound:

1. `PoolPlugin` had no `Drop` implementation to synchronize refcount decrements. When dropped on different threads, unsynchronized `Rc` refcount modifications could lead to data races.
2. The pool used `Rc::strong_count == 1` to track availability, but had no control over when clones were created or destroyed. Callers could clone and keep `PoolPlugin` instances, breaking the availability tracking.
3. The `PoolPlugin` type didn't leverage the type system to enforce safe usage patterns; it relied entirely on external locking that the type system couldn't verify.

This replaces the unsafe implementation with proper thread-safe primitives and move semantics inspired by the deadpool crate.

Changes:

- Replace shared `Rc<RefCell<Plugin>>` with move semantics: plugins are moved out of a `VecDeque` when checked out and moved back on drop
- `PoolPlugin` now owns the plugin and implements `Drop` to automatically return it, making checkout/return atomic and type-safe
- Add `unsafe impl Send/Sync` for `UserDataHandle` to complete the chain of unsafe assertions. This trusts that the existing `unsafe impl<T> Send for UserData<T>` is sound (verified: `UserData::get()` only returns `Arc<Mutex<T>>`, never extracts `T`, so it cannot be moved across threads unsafely)
- Replace busy-wait polling with `Condvar` for efficient waiting
- `PoolPlugin` implements `Deref`/`DerefMut` for ergonomic access
- Use `Weak` reference to pool so checked-out plugins don't keep pool alive
- Fix `function_exists` to handle timeout gracefully instead of panicking
- Add test for pools with captured `PluginBuilder` containing custom functions

Breaking API changes:

- `PoolPlugin` no longer has a `.plugin()` method - it now implements
  `Deref`/`DerefMut`, so you can use it directly where you previously
  called `.plugin()`. Example: `pool.get()?.call(...)` instead of
  `pool.get()?.plugin().call(...)`
- `PoolPlugin::call()` now takes `&mut self` instead of `&self` (direct
  ownership instead of interior mutability via `RefCell`)

Fixes #891

### DIFF
--- a/convert-macros/tests/ui/invalid-encoding.stderr
+++ b/convert-macros/tests/ui/invalid-encoding.stderr
@@ -1,6 +1,7 @@
 error: encoding needs to be specified
 
          = try: `#[encoding(Json)]`
+
  --> tests/ui/invalid-encoding.rs:3:10
   |
 3 | #[derive(ToBytes)]
@@ -12,6 +13,7 @@ error: expected attribute arguments in parentheses: #[encoding(...)]
 
          = note: expects a path
          = try: `#[encoding(Json)]`
+
  --> tests/ui/invalid-encoding.rs:7:3
   |
 7 | #[encoding]
@@ -21,6 +23,7 @@ error: expected parentheses: #[encoding(...)]
 
          = note: expects a path
          = try: `#[encoding(Json)]`
+
   --> tests/ui/invalid-encoding.rs:11:12
    |
 11 | #[encoding = "string"]
@@ -30,6 +33,7 @@ error: unexpected token
 
          = note: expects a path
          = try: `#[encoding(Json)]`
+
   --> tests/ui/invalid-encoding.rs:15:21
    |
 15 | #[encoding(something, else)]
@@ -38,6 +42,7 @@ error: unexpected token
 error: only one encoding can be specified
 
          = try: remove `#[encoding(Encodings)]`
+
   --> tests/ui/invalid-encoding.rs:20:1
    |
 20 | #[encoding(Encodings)]
--- a/runtime/src/function.rs
+++ b/runtime/src/function.rs
@@ -80,7 +80,7 @@ pub(crate) enum UserDataHandle {
     #[allow(dead_code)]
     C(Arc<CPtr>),
     #[allow(dead_code)]
-    Rust(Arc<std::sync::Mutex<dyn std::any::Any>>),
+    Rust(Arc<std::sync::Mutex<dyn std::any::Any + Send>>),
 }
 
 /// UserData is used to store additional data that gets passed into host function callbacks
@@ -94,7 +94,7 @@ pub enum UserData<T: Sized> {
     Rust(Arc<std::sync::Mutex<T>>),
 }
 
-impl<T: Default> Default for UserData<T> {
+impl<T: Default + Send> Default for UserData<T> {
     fn default() -> Self {
         UserData::new(T::default())
     }
@@ -133,7 +133,10 @@ impl<T> UserData<T> {
     /// Create a new `UserData` from a Rust value
     ///
     /// This will wrap the provided value in a reference-counted mutex
-    pub fn new(x: T) -> Self {
+    pub fn new(x: T) -> Self
+    where
+        T: Send,
+    {
         let data = Arc::new(std::sync::Mutex::new(x));
         UserData::Rust(data)
     }
@@ -158,8 +161,8 @@ impl Drop for CPtr {
     }
 }
 
-unsafe impl<T> Send for UserData<T> {}
-unsafe impl<T> Sync for UserData<T> {}
+unsafe impl<T: Send> Send for UserData<T> {}
+unsafe impl<T: Send> Sync for UserData<T> {}
 unsafe impl Send for CPtr {}
 unsafe impl Sync for CPtr {}
 
@@ -188,7 +191,7 @@ pub struct Function {
 
 impl Function {
     /// Create a new host function
-    pub fn new<T: 'static, F>(
+    pub fn new<T: 'static + Send, F>(
         name: impl Into<String>,
         params: impl IntoIterator<Item = ValType>,
         results: impl IntoIterator<Item = ValType>,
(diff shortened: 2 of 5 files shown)


nilslice:
Thanks @bgmerrell! I think both this and #892 are solid. It does seem like #892 may be a slightly safer implementation, and if we're rolling out a fix for that reason maybe we go with that option. 

Does that work for you? Let me know if you've leaned more in one direction or the other since opening these PRs.

bgmerrell:
@nilslice #892 is safer from a code safety perspective for sure because it doesn't use `unsafe` and it enforces that `Send` bound on `UserData`. I just wasn't sure if that new bound was OK. It works for my use case, but there seems to be a pattern in Extism of doing `unsafe` impls for `Send` and `Sync` for various types, and I'm afraid I don't know why that is (I can only assume it's for some added type flexibility):

```
bgmerrell@oberon: ~/code/github.com/extism/extism$ rg "unsafe" |rg 'Send|Sync'
runtime/src/pool.rs:unsafe impl Send for PoolInner {}
runtime/src/pool.rs:unsafe impl Sync for PoolInner {}
runtime/src/pool.rs:unsafe impl Send for Pool {}
runtime/src/pool.rs:unsafe impl Sync for Pool {}
runtime/src/lib.rs:unsafe impl<F: Clone + Fn(&str)> Send for LogFunction<F> {}
runtime/src/lib.rs:unsafe impl<F: Clone + Fn(&str)> Sync for LogFunction<F> {}
[... several more ...]
```

So I created this PR which follows this pattern. Either PR works for me, and I've been running my unit and integration tests for my project with a patched Extism with this PR here for a couple of weeks.

nilslice:
> [...] and it enforces that Send bound on UserData. I just wasn't sure if that new bound was OK.

Yes, thanks for calling this out. I think originally, we planned for mostly non-Rust embedders via FFI, in which case there was no safety enforcement possible. Rust has gained a lot of usage since then though so it's worth tightening up places that we didn't give as much attention to. 

I think #892 is the way to go then. I'll let that thought simmer for another day though, and then close this PR and merge that one. Will ping you with any more questions :)

bgmerrell:
@nilslice just a polite little bump on this ;) I'd like to ditch my extism fork as soon as possible

nilslice (approved):

nilslice:
Thank you @bgmerrell!

nilslice:
Release v1.20.0 is on its way out

bgmerrell:
@nilslice Thanks! You decided on this one over https://github.com/extism/extism/pull/892? Just wanted to be sure you merged what you intended because last discussion it seemed like you were leaning toward https://github.com/extism/extism/pull/892.

nilslice:
ah.. my bad. I took your comment here as indicating this was your expected version to merge. 

I'll probably revert and release #892 


Doing too much at once today 😢

bgmerrell:
@nilslice ah, yeah, sorry it was kind of awkward having two PRs, so I should have put some more context in my bump comment.

If I can make things easier on you by just rebasing https://github.com/extism/extism/pull/892, then I'm happy to do that.

bgmerrell:
@nilslice was thinking about this more this evening.. it's still a net positive that this one landed. It fixes a real bug and it's consistent with the other Extism code. It also let's me get rid of my fork (thanks for the release!).

It's also what I ended up using that last several weeks in my local testing/development because I thought it would be the more likely to be merged... turns out I was right! Just for an unexpected reason! ;)

nilslice:
@bgmerrell - well that's good to hear! since you've exercised the implementation all this time, I am confident in it and if it's all the same to you, will just keep it as-is. 

we have the alternative in the event things should change and can always revisit the decision then.

PR #900, opened 2026-04-14
Add Plugin::memory_status() to query memory limiter state

gkmauer:
Add a public memory_status method to Plugin that returns the current memory limiter values (bytes_left and max_bytes), following the same read-only query pattern as fuel_consumed. Returns None when no memory limit is configured.

- Add MemoryStatus struct with Clone, Copy, Debug derives
- Make MemoryLimiter fields pub(crate) for internal access
- Re-export MemoryStatus from crate root
- Add unit tests for None/Some cases and trait derivations

### DIFF
--- a/runtime/src/current_plugin.rs
+++ b/runtime/src/current_plugin.rs
@@ -24,8 +24,17 @@ pub struct CurrentPlugin {
 unsafe impl Send for CurrentPlugin {}
 
 pub(crate) struct MemoryLimiter {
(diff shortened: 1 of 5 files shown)


nilslice:
@gkmauer - sorry for the delay on this. I have not thought enough about whether this is something we want to include as I have not confirmed it is possible to implement across the other Host engines (wazero and web/nodejs). 

Have not forgotten, just wanted to ping.

PR #902, opened 2026-05-05
Remove dependency on wiggle

oscargus:
Using the default feature for wiggle leads to witx being pulled in. witx in unmaintained and has some quite old dependencies.

It seems like it is OK to skip this for extism. (As in the only failing test locally, also fails with default features...)

Edit: then wiggle is actually not used explicitly in the code base and I get the same test failure if I remove it...

Edit: second commit removes wiggle completely.

Edit: wiggle (and witx) is still pulled in through wasi-common, but at least it is a bit of a cleanup for extism to get rid of it there...

Suggested to squash the commits. Just kept them for you to select which variant should be there.

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -27,8 +27,8 @@ anyhow = "1"
 async-trait = "0.1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-toml = "0.9"
-sha2 = "0.10"
+toml = "1"
+sha2 = "0.11"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3.23", features = [
   "std",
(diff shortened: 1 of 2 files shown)


nilslice:
@oscargus I'd be happy to merge this, but there seem to be conflicts now (perhaps due to #904 which I merged ahead of this based on what seemed like higher priority). 

If you'd like to resolve and update the branch I can get it into a release by the end of the weekend or early next week. 

Thanks!

oscargus:
I've updated the PR (and squashed the commits).

oscargus:
Not sure if the test failure is related, but a bit surprising.

oscargus:
Closing this as I cannot get the tests to pass, even with wiggle...

PR #904, opened 2026-06-07
Add anyhow to wasmtime to enable compilation with no-default-fea…

### LINKED ISSUE #903: 1.30 breaks when using anyhow(?): `std::error::Error` is not implemented for `wasmtime::Error`
Just tried to update to 1.30 and ended up with lots of these errors:
```
error[E0277]: `?` couldn't convert the error: `wasmtime::Error: std::error::Error` is not satisfied
   --> C:\Users\Oscar\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\extism-1.30.0\src\current_plugin.rs:209:50
    |
209 |         if let Some(d) = xs.data_mut(&mut *store)? {
    |                             ---------------------^ the trait `std::error::Error` is not implemented for `wasmtime::Error`
    |                             |
    |                             this has type `Result<_, wasmtime::Error>`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = note: required for `extism_convert::Error` to implement `From<wasmtime::Error>`

error[E0277]: `?` couldn't convert the error: `wasmtime::Error: std::error::Error` is not satisfied
   --> C:\Users\Oscar\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\extism-1.30.0\src\current_plugin.rs:238:14
    |
238 |               )?;
    |                ^ the trait `std::error::Error` is not implemented for `wasmtime::Error`
    |
   ::: C:\Users\Oscar\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\extism-1.30.0\src\lib.rs:5:33
    |
  5 |       ($store: expr, $x:expr) => {{
    |  _________________________________-
  6 | |         let y = $x;
  7 | |         if y.is_err() && $store.get_fuel().is_ok_and(|x| x == 0) {
  8 | |             Err(wasmtime::Error::msg("plugin ran out of fuel"))
...   |
 12 | |     }};
    | |_____- this has type `Result<_, wasmtime::Error>`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = note: required for `extism_convert::Error` to implement `From<wasmtime::Error>`
```

My guess is that it relates to #901 @zoosky 

I depend on extism with `"default-features" = false`, but the tests pass with this so not really sure what happens.

I note that `anyhow` is no longer imported in `wasmtime` (from `Cargo-lock`), so I guess that maybe there should be an `anyhow` feature in `extism` that enables `wasmtime/anyhow` or add the `anyhow` feature in `wasmtime` by default.

Edit: it is possible to run with default features or add the feature `wasmtime-default-features`, but that also adds quite a bit of additional dependencies.

oscargus:
…tures and anyhow

Closes #903 

(When writing the bug report, I also figured out how to solve it etc. And I guess it makes sense to have a proper issue as well...)

Edit: one can argue if this should be default or not.

Edit: turns out that it didn't work to add as a feature since the all-features-tests were failing then. So just added it among the wasmtime features.

Edit: all-fetures-tests fail now as well. Not sure it is related to my change and if it is, I have no idea how to fix it...

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -10,7 +10,7 @@ version.workspace = true
 
 [dependencies]
 wasmtime = { version = "43", default-features = false, features = [
-  # 'anyhow',
+  'anyhow',
   'cache',
   'gc',
   'gc-drc',
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -10,6 +10,7 @@ version.workspace = true
 
 [dependencies]
 wasmtime = { version = "43", default-features = false, features = [
+  # 'anyhow',
   'cache',
   'gc',
   'gc-drc',


nilslice:
Thanks, but it looks like CI is failing on something.

oscargus:
Turns out that the CI failure is not related to this (or rather it happens anyway, there can be issues with this as well, but it is not clear).

The first commit now adds a commented out feature (to trigger the CI) and it also fails.

PR #906, opened 2026-07-05
Share variables across plugin instances.

milesj:
@nilslice 

I was testing a pooled plugin system here: https://github.com/moonrepo/proto/pull/1043

And noticed that vars (via `var::`) are not shared across plugin instances. This kind of feels like an oversight, so thought I would take a stab at fixing it.

### DIFF
--- a/runtime/src/current_plugin.rs
+++ b/runtime/src/current_plugin.rs
@@ -2,11 +2,16 @@ use wasmtime::error::Context as _;
 
 use crate::*;
 
+/// Plugin variables, shared across every instance created from the same
+/// [`CompiledPlugin`]. Wrapped in an `Arc<Mutex<..>>` so that updates made by
+/// one instance are visible to all others and survive a store reset.
+pub(crate) type Vars = std::sync::Arc<std::sync::Mutex<std::collections::BTreeMap<String, Vec<u8>>>>;
+
 /// CurrentPlugin stores data that is available to the caller in PDK functions, this should
 /// only be accessed from inside a host function
 pub struct CurrentPlugin {
-    /// Plugin variables
-    pub(crate) vars: std::collections::BTreeMap<String, Vec<u8>>,
+    /// Plugin variables, shared across every instance
+    pub(crate) vars: Vars,
 
     /// Extism manifest
     pub(crate) manifest: extism_manifest::Manifest,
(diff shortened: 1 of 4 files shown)


nilslice:
Thanks! But this was actually intentional, as each instance of a plug-in memory is supposed to be fully isolated from any other.

Do you have specific needs that this satisfies?

milesj:
I figured that was probably the case.

I am working on a pooled plugin approach using multiple instances (via compiled plugins) to parallelize better. This has the negative side-effect of `var` caching not working correctly, as each func call may be using a different instance than before.

What about this patch instead? I can then do the var syncing on my end.

```patch
Expose the plugin variable store on `Plugin`

`CurrentPlugin::vars()` / `vars_mut()` are already public, but they are
only reachable from within a host function invocation, since the path
from `Plugin` to `CurrentPlugin` (`Plugin::store` / the `Internal`
trait) is `pub(crate)`.

Hosts that maintain a pool of instances created from one
`CompiledPlugin` need a way to seed and snapshot the variable store
between calls, so that guest `var::get`/`var::set` state (commonly used
as a cross-call cache by plugins) behaves consistently regardless of
which instance serves a call. Exposing the existing accessors on
`Plugin` makes that possible without any behavioral change.

Apply to: runtime/src/plugin.rs (impl Plugin, e.g. after `cancel_handle`)

--- a/runtime/src/plugin.rs
+++ b/runtime/src/plugin.rs
@@ impl Plugin {
     /// Get a `CancelHandle`, which can be used from another thread to cancel a running plugin
     pub fn cancel_handle(&self) -> CancelHandle {
         self.cancel_handle.clone()
     }

+    /// Get a reference to the plugin's variable store. Variables persist
+    /// across calls on this instance until the store is reset.
+    pub fn vars(&self) -> &std::collections::BTreeMap<String, Vec<u8>> {
+        self.current_plugin().vars()
+    }
+
+    /// Get a mutable reference to the plugin's variable store. This allows
+    /// hosts to seed, share, or snapshot variables between instances, for
+    /// example when pooling multiple instances of a `CompiledPlugin`.
+    ///
+    /// Must not be called while a call is executing on this plugin.
+    pub fn vars_mut(&mut self) -> &mut std::collections::BTreeMap<String, Vec<u8>> {
+        self.current_plugin_mut().vars_mut()
+    }
+
     pub(crate) fn clear_error(&mut self) -> Result<(), Error> {
```

nilslice:
@milesj - I have a different solution in mind here that is part of larger feature addition. if we decided to instead give a (configurable) durable backing store as part of extism builtins (e.g. a sqlite db, with swappable persistence - memory, disk, cloud - set by the Host), you could establish any pattern you want for plugins to store/load values across plugin execution. 

curious for your thoughts on this!

milesj:
It seems like that could work.

PR #908, opened 2026-07-15
fix(ci): publish extism_sys wheels after the build finishes

### LINKED ISSUE #885: extism-sys==1.13.0  only has a wheel for MacOS?
Is this intentional? UV doesn't seem to be able to install extism without special config, even:

```toml
[tool.uv]
required-environments=[
    "sys_platform == 'linux' and platform_machine == 'x86_64'",
    "sys_platform == 'linux' and platform_machine == 'aarch64'",
    ]
```

doesn't seem to work, but explicitly adding extism-sys<1.13.0 does.

jack-davies:
Hey @nilslice and team, thanks for an excellent project!

Debugging an issue in our CI I noticed there are quite a few missing wheels in pypi across many releases/architectures of **extism_sys**.

It appears there is a race condition between `release.yml` and `release-python.yaml`. `release-python.yaml` is triggered as soon as the GitHub release is created, but before all wheels have been built and uploaded to the GitHub release. Only those wheels that have been built and uploaded to the GitHub release by the time `release-python.yaml` hits the pypi upload step land successfully.

The minimal fix here is to trigger `release-python.yaml` only upon successful completion of `release.yml`.

Resolves https://github.com/extism/extism/issues/885.

### DIFF
--- a/.github/workflows/release-python.yaml
+++ b/.github/workflows/release-python.yaml
@@ -1,8 +1,7 @@
 name: Release Python SDK
 
 on:
-  release:
-    types: [published, edited]
+  workflow_call:
 
 jobs:
   release-sdks:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -227,3 +227,9 @@ jobs:
             *.txt
             *.whl
         if: github.ref == 'refs/heads/main'
+
+  release-python:
+    needs: [release]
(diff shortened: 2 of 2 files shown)


abowcut:
Been waiting ages for this, thanks @jack-davies !

nilslice:
Thank you! Taking a look

nilslice (approved):

PR #912, opened 2026-09-01
Upgrade to Wasmtime 48 (LTS)

bgmerrell:
## Motivation

Extism currently depends on a Wasmtime release line that does not receive CVE
or RustSec backports.

This upgrades Extism to Wasmtime 48, an LTS release, so Extism can consume
security fixes backported to a supported Wasmtime release rather than requiring
a full runtime upgrade for each security issue.

Moving to Wasmtime 48 requires updates to Extism's Rust toolchain, WASI Preview
1 integration, and fuel handling. Those required and related changes are
described below.

## Wasmtime 48 LTS

Upgrade the runtime from Wasmtime 43 to Wasmtime 48.

This is the primary purpose of the PR: move Extism onto a Wasmtime release line
that receives LTS security backports.

The Rust toolchain is updated from 1.94.0 to 1.95.0 because Rust 1.95 is the
minimum version required by Wasmtime 48.

## WASI Preview 1 migration

Wasmtime 48 no longer publishes the `wasi-common` API used by Extism. The
supported Preview 1 integration is provided by `wasmtime-wasi`, so this PR:

- replaces `wasi-common` with `wasmtime-wasi`;
- enables the `p1` feature without enabling unrelated default features;
- replaces `wasi_common::WasiCtx` with
  `wasmtime_wasi::p1::WasiP1Ctx`;
- uses `wasmtime_wasi::p1::add_to_linker_sync`;
- uses `WasiCtxBuilder` to configure preopened directories and inherited
  stdout/stderr; and
- updates WASI exit handling to use `wasmtime_wasi::I32Exit`.

This is migration work required to retain Extism's existing synchronous WASI
Preview 1 behavior on Wasmtime 48. It does not move Extism plugins to the
Component Model or WASI Preview 2.

### Read-only preopened directories

The new `wasmtime-wasi` API provides native read-only and read-write filesystem
permissions through `FsPerms`.

Extism previously implemented read-only preopens with a custom
`wasi_common::WasiDir` wrapper. That wrapper depended on the API being removed,
so this PR replaces it with:

```rust
FsPerms::ReadOnly
FsPerms::ReadWrite
```

This preserves the existing `ro:` allowed-path behavior while relying on
Wasmtime's supported filesystem permission implementation. The custom
`ReadOnlyDir` implementation and its `async-trait` dependency are no longer
needed.

## Fuel accounting

Wasmtime 48 meters WebAssembly executed during module instantiation and
initialization. This makes initialization a meaningful part of fuel accounting:
a plugin can exhaust fuel before its requested export begins, and an unbounded
initialization allowance would allow a start function, constructor, or reactor
initializer to run indefinitely.

At the same time, the existing `with_fuel_limit` API is understood as the fuel
available to each plugin call. Allowing implementation-dependent initialization
costs to consume that call budget would make the effective call allowance vary
between the first and later calls.

This PR therefore introduces a distinct initialization budget while preserving
the existing API defaults.

### Default behavior

Existing callers do not need to change their configuration.

```rust
PluginBuilder::new(wasm)
    .with_fuel_limit(call_fuel)
```

When only `with_fuel_limit` is configured:

- initialization receives a budget equal to the configured call limit; and
- each plugin call receives a fresh call budget.

This keeps initialization bounded without requiring existing hosts to configure
a new option.

If no fuel limit is configured, Wasmtime fuel instrumentation remains disabled
and both initialization and calls remain unmetered, matching existing behavior.

### Initialization override

Hosts that need different budgets can use:

```rust
PluginBuilder::new(wasm)
    .with_fuel_limit(100_000)
    .with_initialization_fuel_limit(10_000)
```

The behavior is:

| Call fuel limit | Initialization override | Initialization | Each call |
| --- | --- | --- | --- |
| None | None | Unmetered | Unmetered |
| `C` | None | `C` | `C` |
| `C` | `I` | `I` | `C` |
| None | `I` | Configuration error | N/A |

By default, initialization inherits the configured call fuel limit.
`with_initialization_fuel_limit` overrides only that inherited initialization
budget; it does not change the fuel available to plugin calls. Requiring a call
limit avoids enabling Wasmtime fuel instrumentation for initialization and then
simulating unmetered calls with `u64::MAX`.

The initialization budget covers WebAssembly executed during:

- linking;
- module instantiation;
- initialization functions; and
- constructors.

Once initialization completes, Extism refreshes the normal call budget before
input marshalling. The call budget therefore covers:

- Extism kernel input and output operations;
- the requested exported function; and
- linked WebAssembly executed during that call.

The new initialization override is currently available through the Rust
`PluginBuilder`. This PR does not change the C ABI.

## Documentation

The runtime documentation still described Extism as supporting Wasmtime
versions 27 through 30 and documented cache configuration requirements from
those releases.

That guidance is no longer accurate for Wasmtime 48. This PR:

- removes the stale version range;
- removes the obsolete required `enabled` cache setting;
- links to current, version-neutral cache documentation;
- corrects the documented cache-loading API;
- documents call and initialization fuel behavior; and
- corrects the implementation reference in the
  `with_wasmtime_config` documentation.

Keeping the cache guidance version-neutral should prevent it from becoming
incorrect during the next Wasmtime upgrade.

## Validation

This passes Extism unit tests and unit tests of my application that uses Extism.


nilslice (approved):
this looks great - thank you for taking the careful approach to ensure compatibility and to minimize the wasmtime api dependency. much appreciated!

PR #917, opened 2026-09-17
fix(runtime): fall back to disabled wasmtime cache when no default config path exists

### LINKED ISSUE #851: cache_config not working on Android aarch64
Hello, I was working on my app and noticed that when building a Plugin on an `aarch64` device it would not work and actually throw this error:

`config file not specified and failed to get the default`

After some digging i found out that the issue happened during the creation of the `PluginBuilder`.

```rust
            let plugin_builder =
                PluginBuilder::new(WasmInput::Manifest(plugin_manifest));
            let compiled_plugin = CompiledPlugin::new(plugin_builder)?;
            let plugin = Plugin::new_from_compiled(&compiled_plugin)?;
```

Running this code would fail; same for this code : 

```rust
  let plugin = Plugin::new(&plugin_manifest, [], true).unwrap();
```

And the only way I found to actually "fix" this was to disable the `cache_config` by adding `with_cache_disabled()` to the `PluginBuilder` : 

```rust
            let plugin_builder =
                PluginBuilder::new(WasmInput::Manifest(plugin_manifest)).with_cache_disabled();
            let compiled_plugin = CompiledPlugin::new(plugin_builder)?;
            let plugin = Plugin::new_from_compiled(&compiled_plugin)?;
```

I reckon this issue arises because of PathBuf and how it doesn't work on some android devices.

And this part is not able to get the default config.

```rust
        match &builder.options.cache_config {
            Some(None) => (),
            Some(Some(path)) => {
                config.cache_config_load(path)?;
            }
            None => {
                if let Ok(env) = std::env::var("EXTISM_CACHE_CONFIG") {
                    if !env.is_empty() {
                        config.cache_config_load(&env)?;
                    }
                } else {
                    config.cache_config_load_default()?;
                }
            }
        }

```

Tyagiquamar:
Fixes #851.

Root cause: with default cache settings, CompiledPlugin::new propagates the error from wasmtime::Cache::from_file(None). On platforms where no default config path can be determined (Android), every plugin build fails with 'config file not specified and failed to get the default' unless the user explicitly passes with_cache_disabled().

Change: treat the default cache config as best-effort. If loading it fails, log a warning and continue with the cache disabled instead of failing plugin creation. Explicit paths (with_cache_config / EXTISM_CACHE_CONFIG) still propagate errors.

Verification (rust:1.95-bookworm container, Rust 1.95.0):
- cargo test -p extism --lib test_issue_851 on test-only commit: FAILED (reproduces #851)
- cargo test -p extism --lib test_issue_851 with fix: ok
- cargo test -p extism --lib: 49 passed, 0 failed
- cargo fmt --check: clean
- cargo clippy -p extism --all-features --no-deps -- -D clippy::all: clean


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

